PR #40, opened 2014-09-07
Added --stdout flag to output single file transforms to stdout

citruz:
To use this tool in combination with others, I found it useful to have a option which allows to pipe the output of  a single file transform to stdout instead a file.

Changes:
* Added `output` variable, which can be either file or stdout
* Added `log()` and `logError()` functions so that no debugging messages are written to stdout if `output` is set to `stdout`

Example Usage:

    doctoc README.md --stdout

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ If you want to convert only a specific file, do:
 
     doctoc /path/to/file
 
+To output result to stdout, instead of a file, do:
+
+    doctoc --stdout /path/to/file
+
 #### Example
 
     doctoc README.md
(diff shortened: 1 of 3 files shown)


thlorenz:
Sorry for not responding sooner, I actually like this.
How about we just redirect **all** log messages to `stderr` which is the better way anyways IMO?
So all we'd have to do is change all `console.log` to `console.error`.

citruz:
Sure, you could do that. I seperated the log funtions into `stdout`  and `stderr` because then you can handle the output independently, e.g. redirect regular log messages to `/dev/null` and just get errors on the console (on UNIX of course).

thlorenz:
Yeah, either way works for me.
Could you please squash these commits into one and rebase on master? Currently there are merge conflicts.

citruz:
Replaced by #48

PR #41, opened 2014-09-23
Add gitlab mode support

jingege:
Add gitlab mode support

see https://github.com/thlorenz/anchor-markdown-header/pull/10


thlorenz:
This was added in the latest release. Although I still would like to see a test for this (I just added the flag and updated the docs).

PR #45, opened 2015-01-16
allowing user to easily specify location of toc in document

AndrewSouthpaw:
Doctoc already has functionality to specify location of TOC, but it's not called out directly. Added this to the README, as well as a few other minor additions.

Adjusted regexp for doctoc TOC label to be simply `<!-- START doctoc ` and `<!-- END doctoc ` to make it easier for users to specify where they want the TOC to be inserted.


thlorenz:
Could you indicate that you could also just copy/paste the generated toc anywhere in the readme in order to have it update right in that place?

AndrewSouthpaw:
Added an extra line. Let me know what you think.

thlorenz:
Thanks, published as `0.11`

Issue #46, opened 2015-01-19
Bitbucket nested links


plumlee:
It appears that Bitbucket markdown formatting expects nested lists to have an indention of 4 characters in order to properly nest, according to [their documentation site](https://bitbucket.org/tutorials/markdowndemo/overview#markdown-header-lists). Search that page for 'Lists can be embedded in lists.' - sorry I can't seem to figure out how to get the raw file + the line numbers for a good link.

Doctoc is currently generating an indention level of 2 spaces for each nested list item in the navigation header. Would you be open to a PR that allowed a command line param to adjust this setting?

I don't know when this might have changed, so please clue me in if I'm missing something obvious. Doctoc works like a charm for me otherwise, thanks for a great tool.

plumlee:
Looks like the 4 spaces is their official standard (Bitbucket, I mean) so I think this would be better as the default versus an option. Doing a PR with the change.

PR #47, opened 2015-01-19
Bitbucket mode uses 4 spaces for nested lists.

plumlee:
Implements fix for #46.


plumlee:
Changed to use a variable and pushed, looks like the PR updated automatically as well. I did consider making it more flexible by exposing the list of options in a single module, then doing more coding so you could easily have multiple options that used 4 spaces. If you would prefer that, let me know.

thlorenz:
Thanks, LGTM - don't worry about the list we can do that once we need it.

thlorenz:
Thanks, went out with `0.10`.

PR #48, opened 2015-01-23
Added --stdout flag to output single file transforms to stdout.

citruz:
Added --stdout flag to output single file transforms to stdout, so that it can be used in conjunction with other tools.
Replacement for #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ If you want to convert only a specific file, do:
 
     doctoc /path/to/file
 
+To output result to stdout, instead of a file, do:
+
+    doctoc --stdout /path/to/file
+
 #### Example
 
     doctoc README.md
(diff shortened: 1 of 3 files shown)


thlorenz:
This breaks some logging features using format options like `'%s'`.
Please fix those, i.e. by using the `format` function found in the node core `util` module.

Example output of breakage:

```
➝  node doctoc.js README.md

DocToccing single file "%s" for %s.

==================

"%s" will be updated

Everything is OK.
```

The `%s` should actually be the filename.

citruz:
You're absolutely right, sorry for that. I fixed this and another bug produced by me.

thlorenz:
Any news on this?

thlorenz:
Closing due to inactivity. Please open another PR that takes my suggestions into account if you're still interested.

PR #52, opened 2015-01-28
Custom-TOC-title and max-heading-level options added, argument parsing improved

mklement0:
* You can now specify a custom TOC title with `--title <title>`.
* You can now limit TOC entries to a max. heading level with `--maxlevel <level>`.
  * Note: For backward compatibility, headings extracted from HTML are limited to 4 levels - no such constraint is based on Markdown headings.
* Argument parsing improved by switching to the 'minimist' library.
  * Now supports `-h` and `--help` and reports syntax errors,

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ by github or other sites via a command line flag.
   - [Using doctoc to generate links compatible with other sites](#using-doctoc-to-generate-links-compatible-with-other-sites)
     - [Example](#example-1)
   - [Specifying location of toc](#specifying-location-of-toc)
+  - [Specifying a custom TOC title](#specifying-a-custom-toc-title)
+  - [Specifying a maximum heading level for TOC entries](#specifying-a-maximum-heading-level-for-toc-entries)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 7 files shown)


mklement0:
It's a non-trivial PR, but I've added tests that hopefully cover all the changes.
(Note that all current tests only test at the JS level, not direct invocation of the CLI.)

I've left `--nodejs` in there for now.

I noticed that there was a default cap on the level of headings to include in the TOC, but it only applied to headings extracted from _HTML_, not headings extracted from Markdown text (whether with hash prefix or underlined). For backward compatibility I've retained this _selective_ default value.

Some terminology notes:
* "heading" is probably more precise than "header" in the context of this project; "header" can also mean the piece of text at the top of *every* page (cf. "footer").
* The abbreviation "No." to mean "number" (used in German, via Italian "Numero") is not used in English and is likely to be confused with "no"; therefore it's better to stay away from variable names such as `maxHeaderNo` (`maxHeaderNum` is an alternative).

thlorenz:
Thanks, went out as `0.12`.

mklement0:
Awesome, thanks.

thlorenz:
No problem :)

Also added you as collab, so now you can work in a branch in this repo for future features/fixes.
We'll discuss in there and once we all agree we can merge to master.

You can also close issues now, so I hope you'll be able to help triage those. Especially with the lately added features (most yours) I expect some questions.

mklement0:
Thanks, I appreciate the trust.

Issue #73, opened 2015-07-03
TOC creates invalid links when headers have inline images.


Daenyth:
Example README: https://github.com/Cockatrice/Cockatrice/blob/dfd5c5e050e2abee5c3e2f7b6b0ae0387ced46e8/README.md

jez:
So the root cause of both this issue as well as #69 is that we're not treating headers as Markdown; rather we're treating them as plain text and doing some simple regex transformation on that text.

What we should be doing is parsing headers into something like a syntax tree for Markdown, determining what text is actually displayed vs. hidden per Markdown's rules, and rendering an anchor tag in the ToC accordingly.

/cc @thlorenz

Issue #69, opened 2015-06-16
Hyperlinked heading is linked with heading's hyperlink, not heading anchor


geremyCohen:
Hi, very cool project here! I did notice an issue when using with my README:

I have a heading that is linked (not sure if this is bad form, probably is):

```
##### [CNN](https://www.cnn.com)
```

When doctoc is run against this README, it renders the TOC entry as so:

```
 - [[CNN](https://www.cnn.com)](...urlpathstuff#cnn)
```

So when I client on this TOC entry, it sends me to www.cnn.com, not #cnn

tia
geremy

bkuhlmann:
In case it helps, I'm seeing this issue as well. Take a look at the headings (with links) as defined in the Alias and Function sections of this [README](https://github.com/bkuhlmann/dotfiles/blob/release/README.md). I had to remove most of the auto-generated *Usage* section since headings with links cause a lot of conflicts. This is what the TOC *would* look like if generated by DocToc without any editing:

![bad_toc](https://cloud.githubusercontent.com/assets/50245/8891546/d2739c72-32eb-11e5-9bbc-030e345e407d.png)

Issue #36, opened 2014-05-01
Can't regex process <hX> when X is an integer


bloo:
It works for any other letter besides "h", but breaks when in angle brackets, pared with an integer - whether the brackets are escaped or not.

good:
```
a1
<a1>
\<a1\>
h1
<9h>
```

bad:
```
<h2>
\<h2\>
\<hh2\>
\<h2h\>
```
Here's the stack:
```
/usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:11
      if (new RegExp(x.text[0]).test(line)) {
          ^
SyntaxError: Invalid regular expression: /

* HTML5 container semantics
  * proper use of [**\/: \ at end of pattern
    at new RegExp (<anonymous>)
    at /usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:11:11
    at Array.map (native)
    at addLinenos (/usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:8:18)
    at module.exports (/usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:66:13)
    at transform (/usr/local/lib/node_modules/doctoc/lib/transform.js:127:13)
    at /usr/local/lib/node_modules/doctoc/doctoc.js:26:20
    at Array.map (native)
    at transformAndSave (/usr/local/lib/node_modules/doctoc/doctoc.js:24:6)
    at Object.<anonymous> (/usr/local/lib/node_modules/doctoc/doctoc.js:66:1)
```

RyanZim:
Is this fixed by https://github.com/thlorenz/doctoc/pull/107?

thlorenz:
Should be, closing since no action for over a year.

mkfreeman:
Hi, 

I'm having the same issue while writing README documents that describe HMTL elements (namely headers):

```
DocToccing single file "README.md" for github.com.

==================

/usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:11
      if (new RegExp(x.text[0]).test(line)) {
          ^

SyntaxError: Invalid regular expression: /` element tells the reader that the content inside is a top-level heading.

The end tag is just like the start tag except that it has a forward slash added just after the less-than sign (`/: Unterminated group
    at new RegExp (native)
    at /usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:11:11
    at Array.map (native)
    at addLinenos (/usr/local/lib/node_modules/doctoc/lib/get-html-headers.js:8:18)
[...]

PR #77, opened 2015-08-04
Parse headers as markdown (issues #69 #73)

jmanuel1:
Parse markdown headers to grab the wanted text, even if markdown is embedded within the header. Especially if the header has links or inline images.

See #73 for explanation.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -3,7 +3,8 @@
 var _             = require('underscore')
   , anchor        = require('anchor-markdown-header')
   , updateSection = require('update-section')
-  , getHtmlHeaders = require('./get-html-headers');
+  , getHtmlHeaders = require('./get-html-headers')
+  , md            = require('markdown-to-ast');
 
 var start = '<!-- START doctoc generated TOC please keep comment here to allow auto update -->\n' +
             '<!-- DON\'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->'
@@ -24,62 +25,39 @@ function addAnchor(mode, header) {
   return header;
 }
 
+function isString(y) {
+  return typeof y === 'string';
+}
 
-function getHashedHeaders (lines, maxHeaderLevel) {
-  var inCodeBlock = false
-    , lineno = 0;
 
-  // Turn all headers into '## xxx' even if they were '## xxx ##'
-  function normalize(header) {
-    return header.replace(/[ #]+$/, '');
+function getMarkdownHeaders (lines, maxHeaderLevel) {
+  function extractText (header) {
+    return header.children
+      .filter(function (x) {
+        return x.type !== 'Image';
+      })
+      .map(function (x) {
+        if (x.type === 'Link') {
+          return extractText(x);
+        }
+        return x.raw;
+      })
+      .join('')
   }
 
-  // Find headers of the form '### xxxx xxx xx [###]'
-  return lines
-    .map(function (x, idx) {
-      return { lineno: idx, line: x }
-    })
+  return md.parse(lines.join('\n')).children
     .filter(function (x) {
-      if (x.line.match(/^```/)) {
-        inCodeBlock = !inCodeBlock;
-      }
-      return !inCodeBlock;
+      return x.type === 'Header';
     })
     .map(function (x) {
-      var match = /^(\#{1,8})[ ]*(.+)\r?$/.exec(x.line);
-
-      return match && (!maxHeaderLevel || match[1].length <= maxHeaderLevel)
-        ? { rank :  match[1].length
-          , name :  normalize(match[2])
-          , line :  x.lineno
+      return !maxHeaderLevel || x.depth <= maxHeaderLevel
+        ? { rank :  x.depth
+          , name :  extractText(x)
+          , line :  x.loc.start.line
           }
         : null;
     })
     .filter(notNull)
-}
-
-function getUnderlinedHeaders (lines, maxHeaderLevel) {
-    // Find headers of the form
-    // h1       h2
-    // ==       --
-
-    return lines
-      .map(function (line, index, lines_) {
-        if (index === 0) return null;
-        var rank;
-
-        if (/^==+ *\r?$/.exec(line))      rank = 1;
-        else if (/^--+ *\r?$/.exec(line)) rank = 2;
-        else                              return null;
-
-        return !maxHeaderLevel || (rank <= maxHeaderLevel)
-          ? { rank  :  rank
-            , name  :  lines_[index - 1]
-            , line  :  index - 1
-            }
-          : null;
-      })
-      .filter(notNull)
 }
 
 function countHeaders (headers) {
@@ -136,8 +114,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, tit
   var currentToc = info.hasStart && lines.slice(info.startIdx, info.endIdx).join('\n')
     , linesToToc = getLinesToToc(lines, currentToc, info);
 
-  var headers = getHashedHeaders(linesToToc, maxHeaderLevel)
-    .concat(getUnderlinedHeaders(linesToToc, maxHeaderLevel))
+  var headers = getMarkdownHeaders(linesToToc, maxHeaderLevel)
     .concat(getHtmlHeaders(linesToToc, maxHeaderLevelHtml))
 
   headers.sort(function (a, b) {
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "dependencies": {
     "anchor-markdown-header": "^0.5.4",
     "htmlparser2": "~3.7.1",
+    "markdown-to-ast": "~3.0.5",
     "minimist": "~1.1.0",
     "underscore": ">=1.3.3",
     "update-section": "^0.3.0"


lib/transform.js
```diff
@@ -3,7 +3,8 @@
 var _             = require('underscore')
   , anchor        = require('anchor-markdown-header')
   , updateSection = require('update-section')
-  , getHtmlHeaders = require('./get-html-headers');
+  , getHtmlHeaders = require('./get-html-headers')
+  , md            = require('markdown-to-ast');
```
jez: Just curious (I haven't looked into this too much), what were some of the factors that led to this markdown parser? Seems like there are a number of other options: `marked-ast`, `markdown-ast`, `markdown-to-ast` to name a few. Was this one easier to use, faster, more well-documented...)

lib/transform.js
```diff
@@ -24,64 +25,41 @@ function addAnchor(mode, header) {
...
+function getMarkdownHeaders (lines, maxHeaderLevel) {
+  function extractText (header) {
+    return header.children
+      .filter(function (x) {
+        return x.type !== 'Image';
+      })
+      .map(function (x) {
+        if (x.type === 'Link') {
+          return extractText(x);
+        }
```
jez: > ```javascript
> .map(function (x) {
>   if (x.type === 'Link') {
>     return extractText(x);
>   }
>   return x.raw;
> })
> ```

Wouldn't we want to recurse all the way down to the leaves? i.e.

```javascript
.map(function (x) {
  return x.type === 'Str' ? x.raw : extractText(x);
}
```

If not, what's special about links in particular?

lib/transform.js
```diff
@@ -136,8 +114,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, tit
   var currentToc = info.hasStart && lines.slice(info.startIdx, info.endIdx).join('\n')
     , linesToToc = getLinesToToc(lines, currentToc, info);
 
-  var headers = getHashedHeaders(linesToToc, maxHeaderLevel)
-    .concat(getUnderlinedHeaders(linesToToc, maxHeaderLevel))
+  var headers = getMarkdownHeaders(linesToToc, maxHeaderLevel)
     .concat(getHtmlHeaders(linesToToc, maxHeaderLevelHtml))
```
jez: Do we still need to parse HTML headers ourselves? I would have hoped that `markdown-to-ast` would parse that for us, but maybe not.

If it doesn't, do we want to do anything special to handle the case of `<h1><a href="#">Header</a><h1>`? I'm fine with not supporting this edge case.

jez:
I left a few comments in line, mostly just curious about how things work. I think this PR is really cool!

thlorenz:
Good idea to move away from the original simplistic header parser.
We may still need to parse out headers of html manually afaik.

I'm wondering like @jez if this is the only markdown parser that gives us what we need. Specifically I know that marked supports handling sections like html separately. Mainly I'm just interested with your experiences with markdown-to-ast.

It'd be totally rad if we could fix the html header parser as well. Currently it'll pick up html inside code sections (vs. actual inlined html).
Any thoughts on that?

Overall like this PR as well especially since all tests still pass unmodified.
So good from my end once @jez's concerns were addressed.

package.json
```diff
@@ -18,6 +18,7 @@
   "dependencies": {
     "anchor-markdown-header": "^0.5.4",
     "htmlparser2": "~3.7.1",
+    "markdown-to-ast": "^3.0.5",
```
thlorenz: Please stick to `~` ranges for now.

lib/transform.js
jmanuel1: I saw line numbers are used to order headers, and `markdown-to-ast` gives location info. I search on Google and Github for "markdown parsers" and "markdown ast" and used `markdown-to-ast` because it was simpler to use. I did see other options; `marked-ast` doesn't give locations. As for `markdown-ast`... should I try it?

lib/transform.js
jmanuel1: I'm thinking `code spans`, *emphasis*, and **strong** should be kept. I'm not sure how the parser acts with autolinks.

Should we handle the case there is html in a header?

lib/transform.js
jmanuel1: I just wanted to get the PR out so you could take a look at it. `markdown-to-ast` will give ast nodes to HTML, but it won't parse it.

thlorenz: OK, so I suppose we still have to work on fixing issues like #36.

lib/transform.js
jez: No, I was mostly just curious. `markdown-to-ast` looks fine. I only really asked because I know that `marked` boasts about how fast it is, but It looks like you have to do a bit of work to expose the underlying AST, and if it still doesn't give line numbers like you mention, then it's probably not worth it.

lib/transform.js
jez: Oh hmm, I was thinking none of those should be kept. I guess I'll defer to @thlorenz then.



> Should we handle the case there is html in a header?

I don't think that many people have run into this issue. Mostly people have additional markdown within their headers. We can potentially address this if at some point we decide to solidify our story around "parsing" HTML in the first place.

thlorenz:
LGTM overall. Let's squash this and merge to master.
We can address any other issues in another PR.
Unless you have an objection @jez

jez:
:+1:

thlorenz:
@jmanuel1 could you please squash these into one commit so we can merge to master?

thlorenz:
@jez @jmanuel1 not sure why this became dormant.
I believe we got some related issues that would most likely be fixed when this gets pulled in, i.e. https://github.com/thlorenz/doctoc/issues/86

Could we get this squashed and merged? If neither of you have the time to do it I'll try to get to it this or next week.

jmanuel1:
OK, I've figured out squashing. All squashed!

thlorenz:
Tried to finally merge this but am running into a test failure:

```
➝  node -v
v4.1.2

➝  npm ts

> doctoc@0.15.0 test /Users/thlorenz/dev/js/projects/doctoc
> set -e; for t in test/*.js; do node $t; done

TAP version 13

# given a file that includes html with header tags and maxHeaderLevel 8
ok 1 generates correct toc for non html and html headers

# given a file that includes html with header tags using default maxHeaderLevel
ok 2 generates correct toc for non html and html headers omitting headers larger than maxHeaderLevel

1..2
# tests 2
# pass  2

# ok
TAP version 13

# overwrite existing title
ok 1 generates correct toc for when custom --title was passed

# do not overwrite existing title
ok 2 generates correct toc for when custom title exists in README already

# clobber existing title
ok 3 generates correct toc for when --notitle was passed

1..3
# tests 3
# pass  3

# ok
TAP version 13

# given a file with edge-case header names
ok 1 generates a correct toc when headers are weird

1..1
# tests 1
# pass  1

# ok
TAP version 13
# transforming
ok 1 transforms it
ok 2 generates correct anchors
# transforming
ok 3 transforms it
ok 4 generates correct anchors
# transforming
ok 5 transforms it
ok 6 generates correct anchors
# transforming
ok 7 transforms it
ok 8 generates correct anchors
# transforming
ok 9 transforms it
ok 10 generates correct anchors
# transforming
ok 11 transforms it
ok 12 generates correct anchors
# transforming
ok 13 transforms it
ok 14 generates correct anchors
# transforming
ok 15 transforms it
ok 16 generates correct anchors
# transforming
ok 17 transforms it
ok 18 generates correct anchors
# transforming
ok 19 transforms it
ok 20 generates correct anchors
# transforming
ok 21 transforms it
ok 22 generates correct anchors
# transforming
ok 23 transforms it
ok 24 generates correct anchors
# transforming
ok 25 transforms it
ok 26 generates correct anchors
# transforming when old toc exists
not ok 27 transforms it
  ---
    file:   /Users/thlorenz/dev/js/projects/doctoc/test/transform.js
    line:   242
    column: 5
    stack:
      - |
        getCaller (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-assert.js:418:17)
      - |
        Function.assert (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-assert.js:21:16)
      - |
        Test._testAssert [as ok] (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-test.js:87:16)
      - |
        Test.<anonymous> (/Users/thlorenz/dev/js/projects/doctoc/test/transform.js:242:5)
      - |
        emitNone (events.js:72:20)
      - |
        Test.emit (events.js:166:7)
      - |
        Test.emit (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-test.js:104:8)
      - |
        GlobalHarness.Harness.process (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-harness.js:87:13)
      - |
        doNTCallback0 (node.js:408:9)
      - |
        process._tickCallback (node.js:337:13)
  ...
not ok 28 TypeError: Cannot read property 'split' of undefined
  ---
    type:    TypeError
    message: Cannot read property 'split' of undefined
    code:    ~
    errno:   ~
    file:    /Users/thlorenz/dev/js/projects/doctoc/test/transform.js
    line:    245
    column:  14
    stack:
      - |
        Test.<anonymous> (/Users/thlorenz/dev/js/projects/doctoc/test/transform.js:245:14)
      - |
        emitNone (events.js:72:20)
      - |
        Test.emit (events.js:166:7)
      - |
        Test.emit (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-test.js:104:8)
      - |
        GlobalHarness.Harness.process (/Users/thlorenz/dev/js/projects/doctoc/node_modules/tap/lib/tap-harness.js:87:13)
      - |
        doNTCallback0 (node.js:408:9)
      - |
        process._tickCallback (node.js:337:13)
      - |
        Function.Module.runMain (module.js:477:11)
      - |
        startup (node.js:118:18)
      - |
        node.js:952:3
    thrown:  true
  ...
# transforming
ok 29 transforms it
ok 30 generates correct anchors
# transforming
ok 31 transforms it
ok 32 generates correct anchors

1..32
# tests 32
# pass  30
# fail  2
```

Do you know what's going on here @jmanuel1?

PR #82, opened 2015-08-22
strip code blocks before parsing headers - #38

### LINKED ISSUE #38: skip contents in <pre></pre>
some of my text in \<pre\> \<\\pre\> tag accidentally follows an header syntax, is it possible to skip those

yomed:
I've fixed #38 by stripping out the 4 types of markdown code blocks before parsing headers. In doing this, I removed the previous method which filtered out triple-backticked lines during parsing (was only done for parsing hashed headers). Test case included. Let me know if you'd prefer fixing this in some other way. It's worth noting that badly-formed markdown files will break... but presumably they would break anyway.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -26,8 +26,7 @@ function addAnchor(mode, header) {
 
 
 function getHashedHeaders (lines, maxHeaderLevel) {
-  var inCodeBlock = false
-    , lineno = 0;
+  var lineno = 0;
 
   // Turn all headers into '## xxx' even if they were '## xxx ##'
   function normalize(header) {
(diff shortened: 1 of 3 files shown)


jez:
Interesting, thanks for the PR! There is another PR out right now (#77) that refactors the parser to use an actual Markdown parser. There is some discussion regarding whether we can use this new model to naturally solve the issue of headers being parsed out.

If you're curious, you might want to poke around with it (I personally would be interested to see if it passes the tests you added in this PR, but I'm on the move ATM).

Because of the nature of the other PR, we might not want to merge this one until it lands.

yomed:
That sounds like a much better long term fix for issues like this one. Will take a look at it, thanks.

yomed:
Closing since this was covered in #77. The tests in this PR might still be useful, but I'll leave that for another PR.

Anthropic:
@yomed this fix is still needed, I used the latest version in npm and it still fails #77 fixed nothing for my use case, I applied your changes and now it works, I would really appreciate it it you updated your version to not need the changes to getHashedHeaders which doesn't seem to be there any more, but re-opened to submit the rest!

Every other change you made is valid and works. THANK YOU for this PR, it NEEDS to be merged and released IMO.

Anthropic:
The problem is the getMarkdownHeaders parser is fine but there is also the getHtmlHeaders which still adds the wrong tags.

Anthropic:
I moved it and created PR to apply only to html header parser.

PR #90, opened 2015-12-30
[Docker]: Provide a Dockerfile

prologic:
Provilde a [Dockerfile](https://docs.docker.com/engine/reference/builder/) for easier use in Docker-based enviornments where one _does not_ want to necessarily polluate their system with various NodeJS dependencies just to run this lovely tool1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM alpine
+
+ENTRYPOINT ["doctoc"]
+CMD []
+
+VOLUME /source
+
+RUN apk -U add nodejs && \
+    rm -rf /var/cache/apk/*
+
+WORKDIR /app
+COPY . /app
+
+RUN npm install -g doctoc
+
+WORKDIR /source


prologic:
I use this with a bash script in ``$HOME/bin/doctoc`` as a wrapper:

```#!bash
#!/bin/bash

file=${1}

docker run -i -t --rm -v $(pwd):/source prologic/doctoc "${@}"
```

**Usage:**

```#!bash
doctoc --github README.md
```

prologic:
The resulting image is only ~25MB in size!

```#!bash
$ docker images prologic/doctoc
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
prologic/doctoc     latest              32577b53f77c        About an hour ago   25.53 MB
```

I recommend you publish this in to the [Docker Hub](http://hub.docker.com/) so all can benefit :)

jez:
I suppose it's interesting what some people call polluting; I would call Docker "polluting" compared to the alternative of installing through npm.

I'm a bit hesitant to merge this in. From what I can tell, your Dockerfile is exactly what I would have expected it to be; I'm not suspicious based on the content of the Dockerfile. I'm just not sure the additional baggage of having to push a new image to the Docker Hub every time doctoc is updated is worth the effort, if it's already packaged as a command line utility on npm.

We could host that Dockerfile in this repository nonetheless, but it might be confusing to people which was the preferred way to use doctoc. At the very least, it'd be nice to have the [usage information][1] you provided in a comment at the top of the Dockerfile.

This is probably a decision for @thlorenz to make; he'd likely be involved with updating the Docker Hub image for new releases.

[1]: https://github.com/thlorenz/doctoc/pull/90#issuecomment-167923549

prologic:
@jez If (*making assumptions here*) you're a NjodeJS developer and loves those sets of tooling and JavaScript as your language/hammer of choice and you came across a great tool that someone else wrote that happened to be written in Python that you had to ``pip install`` "a bunch of dependencies" you'd be inclinced to agree with my point of view and see that Docker can be used as a tool to run applications in an isolated way without polluting the rest of the system.

That being said; to address your concerns of "effort" required to maintain an up-to-date Docker image on the Docker Hub; this is in fact trivial as when it's setup as an Automated Build; every Git push automatically rebuilds the image.

The idea if putting Docker-specific usage information in the Docker Hub Description/README however is good; we should do that definitely.

jez:
@prologic I understand your viewpoint.

I just looked at Docker's [Automated Builds from GitHub][1] feature. It actually looks pretty cool and has minimal setup. With this feature set up and maybe a note added to the README, I think all my concerns against including this have been addressed.

[1]: https://docs.docker.com/docker-hub/github/

prologic:
:+1:

thlorenz:
LGTM let's merge this to master ... no new version on npm should be necessary since it's only needed on github for docker.

jez:
@thlorenz Feel free to. Note that the usage information in [this comment][1] never made it into the Dockerfile nor the README.

We had also discussed setting up some sort of [Automated Build from GitHub][2], which looks like it'd need you as the repo owner to set up and access settings.

[1]: https://github.com/thlorenz/doctoc/pull/90#issuecomment-167923549
[2]: https://docs.docker.com/docker-hub/github/

thlorenz:
OK thanks @jez. Will hold off then until it's complete (i.e. some info in the Readme is needed first).

Once that's done and merged I'll work with @prologic to set up the automated build stuff.

rogersachan:
Has any progress been made?

jandradap:
Other container https://hub.docker.com/r/jorgeandrada/doctoc/

thlorenz:
Nice @jandradap .

@prologic given there is now a docker container do we still need this?

Issue #88, opened 2015-11-14
Do not parse code


elfet:
Hi,

I have some code in my readme with &lt;h1&gt; tag:

```
'''html
<div>
  <h1>{{ title }}</h1>
  <p>
    {{ text }}
  </p>
</div>
'''
```
doctoc take `<h1>{{ title }}</h1>`  as title, but it's is code!

pavestru:
+1 Almost the identical issue, including code!

jez:
Closed via #107.

PR #105, opened 2016-05-29
Strip code blocks before html parser

### LINKED ISSUE #38: skip contents in <pre></pre>

Anthropic:
Fix #38 and #88 with @yomed 's #82 added only to html header parser

### DIFF
--- a/lib/get-html-headers.js
+++ b/lib/get-html-headers.js
@@ -35,8 +35,18 @@ function rankify(headers, max) {
     })
 }
 
+// Remove code blocks so that false headers are not parsed
+function stripMDCodeBlocks(content) {
+  return content
+    .replace(/```([\s\S]*?)```/g, '') // must be before single backtick check
+    .replace(/`([\s\S]*?)`/g, '')
+    .replace(/<code>([\s\S]*?)<\/code>/g, '')
+    .replace(/<pre>([\s\S]*?)<\/pre>/g, '');
+}
+
 var go = module.exports = function (lines, maxHeaderLevel) {
   var md = lines.join('\n');
+  var source = stripMDCodeBlocks(md);
   var headers = [], grabbing = null, text = [];
 
   var parser = new htmlparser.Parser({
@@ -60,7 +70,7 @@ var go = module.exports = function (lines, maxHeaderLevel) {
   },
   { decodeEntities: true })
 
-  parser.write(md);
+  parser.write(source);
   parser.end();
 
   headers = addLinenos(lines, headers)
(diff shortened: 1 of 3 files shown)


lib/get-html-headers.js
```diff
@@ -35,8 +35,18 @@ function rankify(headers, max) {
     })
 }
 
+// Remove code blocks so that false headers are not parsed
+function stripMDCodeBlocks(content) {
+  return content
+    .replace(/```([\s\S]*?)```/g, '') // must be before single backtick check
+    .replace(/`([\s\S]*?)`/g, '')
+    .replace(/<code>([\s\S]*?)<\/code>/g, '')
+    .replace(/<pre>([\s\S]*?)<\/pre>/g, '');
```
jez: What's the difference between `[\s\S]` and just `.`?

I'd have thought that "one of either whitespace or non-whitespace" is just "any character".

Anthropic: At a guess it was probably the author's habit given . doesn't catch \n but in this case it already has newline removed so it could be changed back to just . if there is an assumption \n will never ever be passed in.

jez:
The reason why we have not yet merged this pull request is that it doesn't solve the problem of nested headers in Markdown in a robust way. Markdown is very complicated to parse, and can't be adequately parsed with regular expressions. For example, I pushed a test that should pass but that currently doesn't currently [to my fork][pull-105].

Ideally, we'd choose one of the following ways to solve the problem:

- Use our markdown parser (`markdown-to-ast`) to find all code blocks and remove them from the `linesToToc`.
- Use a markdown parser to find all HTML headers (in addition to the Markdown headers that `markdown-to-ast` is currently finding for us)

The second solution is more idea. It once-and-for all solves the problem of figuring out what the headers are. This includes issues where there are, say, headers in block quotes or tables.

The first solution is probably more feasible, given the way that `markdown-to-ast` works. I haven't looked into the library in a while, but I remember looking last time for a feature that would pull out HTML headers in addition to Markdown headers and not being able to find such functionality.

Because this is not how we'd like to solve issues #37, #38, and #88, I'm closing this for now. If you care about seeing these issues resolved, I encourage you to look into implementing one of the above methods. Indeed, I plan on doing so tomorrow, so I hope to have these issues wrapped up shortly.

Thanks for your work here! Feel free to use the thread if you have any trailing questions.

[pull-105]: https://github.com/Anthropic/doctoc/compare/master...jez:pull-105

Anthropic:
@jez thanks mate, I misunderstood from the comments in 82, I thought when I noticed getMarkdownHeaders in the code you were referring to that as your proposed fix so when it still didn't work I assumed the getHtmlHeaders was separate and needed an alternative solution!

I have it working for me now with the temporary fix, but I maintain the json-schema-form org on GitHub so I have to devote my dev time to that unfortunately or I would have loved to fix it for you.

Thanks for your time and detailed response, look forward to 1.1 :)

PR #106, opened 2016-05-30
Upgrade markdown-to-ast

### LINKED ISSUE #94: Failed to create TOC
Here is the error:
```
/usr/lib/node_modules/doctoc/node_modules/markdown-to-ast/node_modules/mdast/node_modules/vfile/index.js:537
        throw err;
              ^
46:60: Incorrectly eaten value: please report this warning on http://git.io/vUYWz
```

jez:
Fixes #94.

An upstream bug was affecting our usage of markdown-to-ast. It was fixed
upstream, so we're upgrading our dependency to the latest version.

We also add a minimal test readme that previously failed and which no
longer fails after the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "anchor-markdown-header": "^0.5.4",
     "htmlparser2": "~3.7.1",
-    "markdown-to-ast": "~3.0.5",
+    "markdown-to-ast": "~3.2.3",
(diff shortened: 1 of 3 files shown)


thlorenz:
LGTM, please merge to master and I'll publish another patch.

jez:
Done. Confirmed that tests still pass after rebasing on top of origin/master.

Issue #99, opened 2016-04-04
doctoc failing with https://github.com/iogf/untwisted/blob/master/BOOK.md


iogf:
When i try to generate toc for https://github.com/iogf/untwisted/blob/master/BOOK.md
it fails with html code inside ~~~python code ~~~.

For example, for the above link it gives me: I have highlighed the wrong parts with <---.

    
    
    **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
    
    - [What can i create with untwisted?](#what-can-i-create-with-untwisted)
    - [Untwisted plugins](#untwisted-plugins)
    - [The rapidserv plugin](#the-rapidserv-plugin)
        - [A simple application](#a-simple-application)
        - [The **request** object](#the-request-object)
          - [request.method](#requestmethod)
          - [request.headers](#requestheaders)
          - [request.data](#requestdata)
          - [request.query](#requestquery)
          - [request.path](#requestpath)
          - [request.version](#requestversion)
        - [The basic dir structure](#the-basic-dir-structure)
        - [Quote Application](#quote-application)
          - [quote/static/comment.html](#quotestaticcommenthtml)
          - [quote/templates/show.html](#quotetemplatesshowhtml)
    - [ List of quotes. ](#list-of-quotes)
        - [ {{name}} {{quote}}... ](#name-quote)
    - [ Add quote ](#add-quote)
          - [quote/templates/view.html](#quotetemplatesviewhtml)
    - [ {{name}} ](#name) <--- It shouldnt be here.
        - [ {{quote}} ](#quote) <--- it shouldnt be here.
          - [quote/app.py](#quoteapppy)
          - [Running](#running)
        - [Imup Applicaiton](#imup-applicaiton)
          - [imup/templates/view.jinja](#imuptemplatesviewjinja)
    - [ Upload image ](#upload-image)
    - [ Images ](#images)
        - [ {{index}} ](#index) <--- It shouldnt be here.
    - [ Upload image ](#upload-image-1)
    - [ Images ](#images-1)
        - [ {{index}} ](#index-1)
          - [imup/app.py](#imupapppy)
    - [ Bad request ](#bad-request)
    - [ Bad request ](#bad-request-1)
        - [quickserv script](#quickserv-script)
    - [The requests plugin](#the-requests-plugin)
    - [The event-driven paradigm](#the-event-driven-paradigm)
        - [Skeletons](#skeletons)
[...]

PR #91, opened 2015-12-31
Parse headers as markdown (bugs fixed)

jez:
This pull request is a continuation of #77, as it looks like that PR has
gone stale. It concluded with a note that the tests didn't seem to pass,
so I've fixed the failing tests, made the changes necessary to fix #69
and #73, and added tests verifying that they work as expected.

The root of the failing tests in #77 was that `markdown-to-ast` (which
uses Remark under the hood) was incorrectly parsing this snippet of
Markdown:

    
    ## Header

This came up because we were misusing the `#parse` method of the
thlorenz/update-section library we're using: endIdx is actually an
_inclusive_ index, but we were using it as an exclusive index. (As an
aside, during the process I also discovered a bug in update-section,
which I will report in a separate PR. It does not affect this PR nor
doctoc in general).

I've not squashed the commits intentionally so that you can see what's
been changed between this and #77. When you've reviewed it and it looks
good I'll be sure to do so.

Issue #84, opened 2015-08-28
ignore headers within code blocks 


dpinol:
Hi,
Could the headers within code blocks be ignored? When I run doctoc on input below...

	# hola
	~~~
	# adeu
	~~~

... I get the following, I'd only like to have "hola" at the TOC
~~~
- [hola](#hola)
- [adeu](#adeu)
~~~
thanks

jez:
Yeah, that's an issue we've been aware of for some time. I wouldn't be surprised if #77 fixes it though.

jez:
Closed via #107.

PR #107, opened 2016-05-30
Use markdown + html parsers to detect codeblocks

jez:
**Relevant Issues**

Fixes #37, #38, #84, #88, #99.

**Summary of Changes**

This commit builds upon the markdown-to-ast integration we added in #91,
using markdown-to-ast in `lib/get-html-headers.js`.

We first filter our lines to only look for HTML sections (using the
markdown parser), then in these HTML-only sections, we search for headers,
as long as they're not inside pre tags.

This is accomplished by keeping a stack of tags we've descended through. We
can peek at the top of the stack to see if we're directly under a
`pre` tag, meaning we should ignore everything until we see the closing
`pre`. Otherwise, we can collect text from inside heading tags as normal.

Thus, the core changes in this commit are:

- `grabbing` is a stack now, instead of a string
- `markdown-to-ast` lets us filter for only HTML blocks
- `htmlparser2` lets us check if we're inside a heading *or* pre tag

**Test Plan**

This pull request adds two passing tests:

- one that makes sure we eliminate bad headings nested in pre tags, and
- one that would fail a non-robust, regex-based search for code blocks,
  excluding headings that should have been included

Basically, the first checks for false positives, and the second checks
for false negatives.

### DIFF
--- a/lib/get-html-headers.js
+++ b/lib/get-html-headers.js
@@ -1,6 +1,7 @@
 'use strict';
 
-var htmlparser = require('htmlparser2');
+var htmlparser = require('htmlparser2')
+  , md         = require('markdown-to-ast');
 
 function addLinenos(lines, headers) {
   var current = 0, line;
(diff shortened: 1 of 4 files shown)


thlorenz:
Awesome work - fixes so many issues :) !
LGTM just hoping that the markdown to ast part doesn't affect performance especially for large Readmes too much. Have you done a quick `time ...` test of this branch against master?

However correctness is preferred either way (just interested how big the hit is), so merge to master and I'll publish a new version.

Thanks.

jez:
Good call. Not sure what the best way to test for regressions is, but it seems to still do fine on the test suite and on one ~600 line [README from one of my other projects][1]:

```zsh
# origin/master
npm test &> /dev/null  1.55s user 0.20s system 104% cpu 1.661 total
doctoc.js test/fixtures/cmugpi.github.io-README.md  0.30s user 0.04s system 96% cpu 0.352 total

# jez/fiz/headers-in-codeblocks
npm test &> /dev/null  1.61s user 0.20s system 104% cpu 1.726 total
doctoc.js test/fixtures/cmugpi.github.io-README.md  0.42s user 0.05s system 102% cpu 0.453 total
```

| | npm test | cmugpi |
| --- | --- | --- |
| origin/master | 1.661s | 0.352s |
| jez/fix/headers-in-codeblocks | 1.726 | 0.453s |

The test suite includes a lot of small tests, as well as a couple a couple longer fixture files. The extra README I found is ~2.5x longer than any of the fixtures. We're not going up by more than 0.1s in either case. If we really cared about performance, we could probably markdown parse the file once (we're doing it twice right now).

[1]: https://github.com/cmugpi/cmugpi.github.io/blob/develop/README.md

thlorenz:
Thanks published as minor upgrade (since some behavior may have changed).

Issue #108, opened 2016-06-10
Custom marker style


kachkaev:
I'm linting my `*.md` with remark-linter and one rule I'm checking the files against is that the markers should be `*` rather than `-` (markdown equally supports both of these options).

It would be nice if I could tell doctoc which of the two marker symbols to use in order to avoid linter errors.

jez:
I agree! I've got a number of bug-related issues on my plate before I can consider feature requests, but if you really want you're welcome to open a PR (with tests :D).

kachkaev:
Hi @jez! I'm afraid I won't be able to send a PR – not enough time. I just came across your project and tried it on a small app, but decided not to use it for now. I'm keeping an eye on `doctoc` and will be happy to `npm install` it again when there's a real need. What you've done looks useful indeed!

sheriferson:
Hi @jez, I'd like to try and implement this feature (pass an argument/flag to `doctoc` to use `*` instead of `-`), but I'm a bit new to js. What's the best way to run the included tests on my local machine? (running OSX if it makes a difference but my guess is that it doesn't).

jez:
On my phone right now but

npm test

should work if I recall correctly.

sheriferson:
That worked, thanks. Simple enough!

jez:
The fix for this is on `master`, but has yet to be released in an npm version. Feel free to pull doctoc in using git in your package.json if you want the features from the `master` branch.

PR #109, opened 2016-06-11
Adds --asterisk or -a flag to use '*' instead of '-' Re: #108

sheriferson:
With this change, a user can pass `-a` or `--asterisk` to `doctoc` and the generated toc will use `* ` instead of `- ` to prefix the toc entries (as discussed in #108). I added a test for this use case, and also tested it on my own Markdown documents.

I'm happy to make any changes as needed.


test/transform.js
```diff
@@ -327,3 +327,28 @@ check(
...
+    ,     '    * [Method One](#method-one)\n'
+    ,     '    * [Method Two](#method-two)\n'
+    ,         '      * [Main Usage](#main-usage)\n\n\n'
+    ].join('')
+  , undefined
+  , undefined
+  , undefined
+  , undefined
+  , true // set asteriskFlag which is passed to transform() to true
+)
```
jez: Looks like your editor isn't configured to add newlines at the end of the file on save. See "[Why should text files end with a newline?][1]" for more information. 

[1]: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

jez:
Hey thanks for the quick turn around on this! See my inline comments for specific nits.

More generally, how do you feel about making the option a little bit more powerful? What if we could do this:

```shell
# Override default '-' with '*'
doctoc --bullet '*'

# Or use numbered lists
doctoc --bullet '1.'

# Or use GitHub-style check lists
doctoc --bullet `- [ ]`
```

We could have this option default to `-`, as opposed to how the current option defaults to `false`. I'm open to suggestions on what the flag should be called; `bullet` isn't quite right because `1.` isn't a bullet per sé.

test/transform.js
sheriferson: Thanks for the note and explanatory link. I'm about to add a commit that will add the newline, and will respond to the rest of suggestions below.

sheriferson:
I like the idea of making it more powerful.
How about `--prefix` or `--tocprefix`?

As an alternative, an option that doesn't allow for open-ended prefixes can be:

```bash
doctoc --prefix asterisk
doctoc --prefix numbered
doctoc --prefix githubchecklist
```

**Pro**
- Handling the limited options internally will be easier, and we won't have to worry about things like detecting when the user wants a numbered vs bulleted list based on their input, and doing a lot of sanitizing and protecting against weird or messy inputs.

**Con**
- Not the most powerful.

I like this option better, but let me know what you think. I think there are a few commonly used list prefixes, and we can cover the "standard" ones.

In the meantime, I'll read the code some more and see if I can understand how to implement something like the numbered list.

jez:
> **@sheriferson** writes:
> and we won't have to worry about things like detecting when the user wants a numbered vs bulleted list based on their input

This would "Just Work" work as I described it above. Consider the following markdown (note how everything uses `1. ` as a prefix:

```markdown
1. Introduction
  1. Heading 2
    1. Heading 3
  1. Another Heading 2
1. Conclusion
  1. Heading 2 under conclusion
```

This generates the following rich text:

> 1. Introduction
>   1. Heading 2
>     1. Heading 3
>   1. Another Heading 2
> 1. Conclusion
>   1. Heading 2 under conclusion

Markdown is smart enough to do the auto-numbering for you, even at the correct nesting levels. This is why I'm in favor of passing a string option. The options "hyphen", "asterisk", "numbered", and "checklist" are all the ones I can think of right now, but different Markdown parsers might have their own, and I don't want to limit myself to just what I know.

> In the meantime, I'll read the code some more and see if I can understand how to implement something like the numbered list.

Correct me if I'm wrong, but I think implementing this option is a straight forward extension of your current diff:

1. Rename `--asterisk` to `--entryprefix`
1. Change `transform()` to take an `entryprefix` argument from `doctoc.js`, and
1. [remove these lines][1] which currently initialize the entry prefix

[1]: https://github.com/thlorenz/doctoc/pull/109/files#diff-9000d2ea35a5fd71d0a8caffed685d47R138

sheriferson:
You're right. The Markdown conversion works out fine if all entries are prefixed with `1.`
I was considering that the raw text/Markdown toc wouldn't be useful, but if that's not a concern, then it's as easy as you mentioned.

I've already made the changes but am running into one issue: when I try the GitHub checklist option:

```bash
doctoc --entryprefix '- [ ]'
```

I run into the issue that the argument parser interprets the `- [ ]` as an unknown option and displays the unknown option error. Using different surrounds like ` or " doesn't get around that problem. I tried an online search but couldn't find anything. Is there an easy way to get around this that I don't know?

jez:
I imagine it will probably work if you use `* [ ]` instead, correct?

If you wrap up everything else and push it, I can poke around with it on my machine. It's possible that `minimist` (the library we're using for option parsing) has a way to handle stuff like this.

jez:
What happens when you do this?

```
doctoc --entryprefix='- [ ]'
                    ^
            note this equals sign
```

sheriferson:
> I imagine it will probably work if you use * [ ] instead, correct?

Yes it does, but would GitHub interpret `* [ ]` as a checklist as it does `- [ ]`?

Testing: 

* [ ] task one
* [ ] task two

Looks like it does.

> If you wrap up everything else and push it, I can poke around with it on my machine. It's possible that minimist (the library we're using for option parsing) has a way to handle stuff like this.

Will do. I'll make sure it's all good and add a couple of tests and push the commit later today.

> What happens when you do this?

```bash
doctoc --entryprefix='- [ ]'
                    ^
            note this equals sign
```

That does work and adds `- [ ]` as the prefix.

sheriferson:
The last commit changes the `--asterisk` flag to `--entryprefix prefix` flag.
I wrote a couple of tests that test single char, double char, and the numbered prefixes.

jez:
Looks great, thanks!

I'm going to merge this now, and we can get @thlorenz to provide any follow up feedback or changes before he publishes a new version on npm.

jez:
Merged in ec0157b. Thanks again!

sheriferson:
My pleasure!
Technical question: how did the change end up being merged? I'm used to a pull request being merged, but something different happened here and I'm curious.

jez:
I added your remote on my local checkout of the repo, fetched the commits you added, then squashed them into one commit so that there was only one commit representing this change. It's that commit which you can find at ec0157b. You'll notice that the commit message differs from any of the commits you created, though all the changes are there.

You can read all about how it's done here: [How to Rebase a Pull Request][1]. You can read more about the [advantages and disadvantages of rebasing a pull request][2], as the camps are somewhat passionately split. @thlorenz decided on this workflow for doctoc, so he might have some more to say about his motivations, but I enjoy it because it makes looking at the [Git Log history][3] much easier, as I explain in that blog post.

[1]: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request
[2]: https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase/
[3]: http://blog.jez.io/2015/01/16/mastering-git-log-for-collaboration/

Issue #100, opened 2016-04-18
Output to stdout


benma:
Hi

It would be great if doctoc could also be used to figure out if the existing table of contents (generated by doctoc) is up to date.

One way would be to add an option to output the updated markdown file to stdout, instead of updating the file. The output could be diffed to the original.

Reason for the request: we have a lot of markdown files in our repo, and it would be nice if our CI could check if the table of contents have been updated.

jez:
Closed by #111.

benma:
Thank you!

PR #111, opened 2016-06-17
Added option for stdout

RichardLitt:
By adding the `-s` or `--stdout` option, you can print to stdout now. Not the most beautiful PR, but I think it is pretty efficient and it works for me. Let me know if I can edit anything!

Related to issue #100

### DIFF
--- a/test/transform-stdout.js
+++ b/test/transform-stdout.js
@@ -1,0 +1,36 @@
+'use strict';
+/*jshint asi: true */
+
+var test = require('tap').test,
+    fs = require('fs'),
+    exec = require("child_process").exec;
+
+test('\nshould print to stdout with --stdout option', function (t) {
+
+    exec('node doctoc.js test/fixtures/readme-with-custom-title.md --stdout', function (error, stdout, stderr) {
+      if (error) {
+        console.error('exec error: ', error);
+        return;
+      }
+      t.deepEqual(stdout
+        , fs.readFileSync(__dirname + '/fixtures/stdout.md', 'utf8')
+        , 'spits out the correct table of contents')
+
+      t.end()
+    })
+})
+
+test('\nshould print to stdout with -s option', function (t) {
+
+    exec('node doctoc.js test/fixtures/readme-with-custom-title.md -s', function (error, stdout, stderr) {
+      if (error) {
+        console.error('exec error: ', error);
+        return;
+      }
+      t.deepEqual(stdout
+        , fs.readFileSync(__dirname + '/fixtures/stdout.md', 'utf8')
+        , 'spits out the correct table of contents')
+
+      t.end()
+    })
+})
--- a/README.md
+++ b/README.md
@@ -132,4 +132,8 @@ By default,
 - no limit is placed on Markdown-formatted headings,
 - whereas headings from embedded HTML are limited to 4 levels.
 
+### Printing to stdout
+
+You can print to stdout by using the `-s` or `--stdout` option.
+
 [ack]: http://beyondgrep.com/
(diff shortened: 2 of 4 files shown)


thlorenz:
Could you please add a test for this?
You could use `child_process` to launch the doctoc cli tool and check for the output, i.e. make sure it prints if `-s,--stdout` is specified but writes to the file otherwise.
Aside from that LGTM

RichardLitt:
@thlorenz That should do it! 

And thank you! I love doctoc, btw; have been using it for a long time now on many different repos. Amazing thing.

test/transform-stdout.js
```diff
@@ -0,0 +1,67 @@
...
+test('\nshould print to stdout with --stdout option', function (t) {
+
+    exec('node doctoc.js test/fixtures/readme-with-custom-title.md --stdout', (error, stdout, stderr) => {
+      if (error) {
+        console.error(`exec error: ${error}`);
+        return;
+      }
+      t.deepEqual(stdout,
+        `
+DocToccing single file "test/fixtures/readme-with-custom-title.md" for github.com.
```
thlorenz: Huge fan of template literals, but those won't run with older node versions and I'd like to keep these running.
If you could please pull these out into files in the [fixtures](https://github.com/thlorenz/doctoc/tree/master/test/fixtures) and read them in instead, we'd highly appreciate it :).
I know it's more work, but then again nothing else in this repo uses non ES5 features, so for a bit I'd like to stay backwards compat.

thlorenz:
@RichardLitt almost there .. thanks for your patience :)

RichardLitt:
Saw that coming, should have done that myself. :) Fixed now.

test/transform-stdout.js
```diff
@@ -0,0 +1,35 @@
...
+test('\nshould print to stdout with --stdout option', function (t) {
+
+    exec('node doctoc.js test/fixtures/readme-with-custom-title.md --stdout', function (error, stdout, stderr) {
+      if (error) {
+        console.error('exec error: ', error);
+        return;
+      }
+      t.deepEqual(stdout
+        , require('fs').readFileSync(__dirname + '/fixtures/stdout.md', 'utf8')
+        , 'spits out the correct table of contents')
```
thlorenz: final nit: please require `fs` at the top of file.

test/transform-stdout.js
```diff
@@ -0,0 +1,35 @@
...
+
+test('\nshould print to stdout with -s option', function (t) {
+
+    exec('node doctoc.js test/fixtures/readme-with-custom-title.md -s', function (error, stdout, stderr) {
+      if (error) {
+        console.error('exec error: ', error);
+        return;
+      }
+      t.deepEqual(stdout
+        , require('fs').readFileSync(__dirname + '/fixtures/stdout.md', 'utf8')
```
thlorenz: same as above

thlorenz:
Granted final nits will be fixed LGTM now!
@jez are you OK with this as well? If so we can merge to master and publish a new minor.

RichardLitt:
Fixed.

thlorenz:
Cool, thanks @RichardLitt! Giving @jez a bit to give his verdict, but we'll most likely merge by EOD.

jez:
LGTM. I've actually secretly wanted this feature for a while :D

jez:
Also, :+1: for tests for `doctoc.js`, not just `transform.js`. We don't
have any of those yet; it's great to see them getting started.

thlorenz:
Thanks, went out as v1.2.0!

RichardLitt:
Woo! Thanks!

PR #112, opened 2016-06-18
Updated help output

RichardLitt:
I reorganized, limited to 80 chars, added `--stdout`, added shortcut options, put a space between the path description to make it more readable, and also fixed ambiguous working in the stdOut function.


RichardLitt:
@thlorenz I want to do another PR to remove the extra text for the `stdOut` option, so that you can pipe it places. Any chance we can merge this first?

thlorenz:
@RichardLitt since this is all related to the output, we could probably combine the two.
So you could just add those commits on top of this one and add to the title/description.

I assume you're gonna log on stderr to not clutter stdout, so stdout would only contain the output markdown right?
That's how I should've done it in the first place probably.

RichardLitt:
@thlorenz How about that? 

Not sure about stderr -- where should that be applied, best? 

Thanks.

thlorenz:
Basically any `console.log` would become `console.error` except when you write markdown to the terminal.

thlorenz:
So to be clear, removing logs when using stdout option isn't a good idea, but to make things pipeable you just log them on stderr so they print when you do `doctoc --stdout > file.md` but all markdown goes into file.txt since stdout is piped by default.

RichardLitt:
Might need some help with this one. Tried to get it working in this change. What do you think?

doctoc.js
```diff
@@ -17,12 +17,12 @@ function cleanPath(path) {
 }
 
 function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, stdOut) {
-  console.log('\n==================\n');
+  (stdOut) ? console.error('\n==================\n') : console.log('\n==================\n')
```
thlorenz: Almost, but ALWAYS log on `console.error` not only if `stdout` is used.
It's the correct thing to do either way when logging (mistake of my past self for not doing so in the first place ;) )

thlorenz:
@RichardLitt commented in one place on the main idea. This applies for all the other places as well.

Ideally you'd just take the original code and do a search replace `%s/console.log/console.error/g` and then make sure you use `console.log` ONLY for the markdown output.

RichardLitt:
@thlorenz Ahh, go tit. Pushed.

RichardLitt:
@thlorenz Ping!

RichardLitt:
@thlorenz Any news on this? I still use this daily, and would love this change to be merged. :)

thlorenz:
It LGTM in general, sorry for the delay.
Could you add some test where you launch doctoc as a child process and check what got printed to stdout and what to stderr? Also please add a section to the readme explaining the `--stdout` option.

RichardLitt:
I was pretty sure I had already done this; didn't https://github.com/thlorenz/doctoc/pull/111 cover this? Why isn't this in the current package?

RichardLitt:
**Update:** It is in the current package, not sure why I hadn't seen it before. Works with a fresh install. So, this is just an update of that. 

@thlorenz Do you still need the extra test? I don't see how that is different from the test I provided in #111 in transform-stdout.js. Happy to look again, though. 

There already is a section on `-s` and `--stdout` in the README; do you want me to add more to it?

RichardLitt:
@thlorenz hey! Just checking old PRs and saw this. What do you think?

RichardLitt:
@thlorenz I'm curious. What have you moved onto? Do you still have a need to use this code, and are you still maintaining it?

thlorenz:
haven't moved on, just busy :)
However I still would like a test ...

Also any clarification as to how things work should be noted in the readme (if there is a section already add to/modify it please)

Thanks ..

Also would like to have other collabs give their OK before we merge this (hopefully they see this).

RichardLitt:
Updated! Let me know if this looks alright.

thlorenz:
Thanks, on first glance looks great.
Will look this over next week in detail ...

kingbangla191:
Thanks, on first glance looks great.
Will look this over next week in detail ...

RichardLitt:
I am not interested in maintaining this PR; I made it so long ago that I don't remember doing it. Good luck if anyone decides to take it on.

PR #114, opened 2016-07-15
Exclude .github from search

vtravieso:
Excludes the .github folder which contains .md templates for issues and pull requests

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -3,7 +3,7 @@ var path  =  require('path')
  ,  _   =  require('underscore');
 
 var markdownExts = ['.md', '.markdown'];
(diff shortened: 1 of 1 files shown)


jez:
Thanks for the PR, but I think we're going to pass on this one. It's better for us to be conservative here; I can imagine that some users may want TOCs in their `.github/` folders.

Luckily, DocToc already has all you need to ignore these yourself. Since you can pass multiple file names to `doctoc` on the command line, you can use something like

```
find . | grep -v '^./.github' | xargs doctoc
```

(You'll have to play around with the filter, but the idea is the same). You can read more about this [in the README][1].

[1]: https://github.com/thlorenz/doctoc#adding-toc-to-individual-files

Issue #80, opened 2015-08-11
Feature Request: Flag to Only Update Files With Comment Block


reggi:
Is there a flag / option to only add a TOC if there's a comment block?

reggi:
I suppose placing a block at the bottom of the file is the only way to ensure it won't get generated when you assign a dir of files.

jez:
I actually have a PR out (#68) that would let you do this with a bit of ack/ag magic. @thlorenz just needs to merge and push an npm update.

reggi:
Does your pr do `<-- doc toc skip --> ` not sure if thats ideal. I think I'd rather not write comments for each doc. I'd like to see an explicit flag.

```
doctoc **/md.js --explicit
```

And only docs that have the 

```
<!-- START doctoc -->
<!-- END doctoc -->
```

or 

```
<!-- START doctoc generated TOC please keep comment here to allow auto update -->
<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->

<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```

will be written to.

jez:
So my PR actually just lets you pass in a list of files to run doctoc on. If you want to exclude specific files, you can add 'DOCTOC SKIP' in comments to the files to exclude and run

```
doctoc $(ack -L 'DOCTOC SKIP')
```

but if you want something like `--explicit`, you can use `-l` (n.b.: lowercase)

```
doctoc $(ack -l 'START doctoc')
```

Of course, you can feel free to use any other program instead of `ack` that lets you search for files and list their names (`ag`, `pt`, `grep`) but `ack` is what I know so you'll have to consult the docs for the other programs.

jez:
Closing as duplicate of #66

reggi:
@jez you just blew my mind, can't wait to give this a try.

keverw:
I'm running `ack -L 'DOCTOC SKIP' | xargs doctoc` like the readme mentions. It's not updating at all.

```shell
ack -L 'DOCTOC SKIP' | xargs doctoc
```

then I get back:

```shel
Unknown option(s): -
Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))

Available modes are:
  --bitbucket	bitbucket.org
  --nodejs	nodejs.org
  --github	github.com
  --gitlab	gitlab.com
[...]

PR #132, opened 2017-06-24
Added flag for only updating files that already contain doctoc

karfau:
This PR allows the use of the flag `--explicit` (or `-e`).
With the flag only files already containing `START doctoc` will get updated.

This adds the feature requested in #80.

As mentioned in the ticket I'm willing to change the name of the flag, please suggest one if the one I chose doesn't work for you.

Open tasks
- [x] add documentation in README
- [x] add tests (not sure how to, after looking at the current test suite)
- [x] incorporate review feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ by github or other sites via a command line flag.
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-
 ## Installation
 
     npm install -g doctoc
@@ -52,7 +51,16 @@ specified](#using-doctoc-to-generate-links-compatible-with-other-sites).
 
 ### Update existing doctoc TOCs effortlessly
 
-If you already have a TOC inserted by doctoc, it will automatically be updated by running the command (rather than inserting a duplicate toc). Doctoc locates the TOC by the `<!-- START doctoc -->` and `<!-- END doctoc -->` comments, so you can also move a generated TOC to any other portion of your document and it will be updated there.
+If you already have a TOC inserted by doctoc, it will automatically be updated 
+by running the command (rather than inserting a duplicate toc).
+Doctoc locates the TOC by the `<!-- START doctoc -->` and `<!-- END doctoc -->` comments, 
+so you can also move a generated TOC to any other portion of your document and it will be updated there.
+
+You can tell doctoc to **only update** files that **already** have a TOC 
+(including those that [specify location of toc](#specifying-location-of-toc)),
+by adding the `--explicit` flag:
+
+    doctoc . --explicit
 
 ### Adding toc to individual files
 
--- a/doctoc.js
+++ b/doctoc.js
@@ -16,13 +16,13 @@ function cleanPath(path) {
   return homeExpanded.replace(/\s/g, '\\ ');
 }
 
-function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, stdOut) {
+function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, stdOut, explicit) {
   console.log('\n==================\n');
 
   var transformed = files
     .map(function (x) {
       var content = fs.readFileSync(x.path, 'utf8')
-        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix);
+        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix, explicit);
       result.path = x.path;
       return result;
     });
@@ -36,9 +36,11 @@ function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPref
     })
   }
 
-  unchanged.forEach(function (x) {
-    console.log('"%s" is up to date', x.path);
-  });
+  if (!explicit) {
+    unchanged.forEach(function (x) {
+      console.log('"%s" is up to date', x.path);
+    });
+  }
 
   changed.forEach(function (x) { 
     if (stdOut) {
@@ -54,7 +56,7 @@ function printUsageAndExit(isErr) {
 
   var outputFunc = isErr ? console.error : console.info;
 
-  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
+  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] [--explicit] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
   outputFunc('\nAvailable modes are:');
   for (var key in modes) {
     outputFunc('  --%s\t%s', key, modes[key]);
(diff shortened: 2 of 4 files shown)


revelt:
This is a good-one, imho doctoc should behave like this by default. How many times I accidentally ran doctoc and got nasty unwanted table of contents at the very top. Personally, I always, 100% of cases, want to control where TOC is put, to control the title and if the placeholder is not there, it's intentional. 👍 👍 👍

karfau:
@revelt I prepared a branch in my fork that has this default behavior: https://github.com/karfau/doctoc/tree/explicit-default that has what you need. I just noticed it is 1 commit behind the main repo, if you need it up to date, just let me know.

revelt:
@karfau thank you sir! 🍻

thlorenz:
I like it. So once you check off the other points (tests + docs) I'd be happy to merge if everyone else is.

karfau:
Glad you like it.

I tried to add something to the README, please have a look if this is what you had in mind.

For the tests, I need some help, as I'm not familiar with `tap` and am not able to understand how/where I would add the tests to the existing structure.
I had a look at the test coverage (by temporarily change the `test` script in `package.json` to `set -e; tap test/*.js --cov --coverage-report=lcov;`) and it showed me that none of the cases where the method `transform` returns `{ transformed: false }` is covered by any test.

I would have expected that I would be able to run all the tests when setting the new last argument to transform to `true` and still all tests would pass, but in this case I have 16 failing tests in `tests/transform.js` with the following message:
```
  transforming
  not ok Cannot read property 'split' of undefined
    stack: >
      Test.<anonymous> (test/transform.js:16:25)
    
      check (test/transform.js:12:3)
    
      Object.<anonymous> (test/transform.js:380:1)
    
      Module.replacementCompile
      (node_modules/nyc/node_modules/append-transform/index.js:58:13)
    
      module.exports
      (node_modules/nyc/node_modules/default-require-extensions/js.js:8:9)
    at:
      line: 16
      column: 25
      file: test/transform.js
      function: Test.<anonymous>
    type: TypeError
    test: transforming
    source: |
      var data = res.data.split('\n');
```
... but maybe this approach was not a valid one?

I'ld love to learn/understand how this testsuite works, or at least what to test and where to add it.

Best Christian

thlorenz:
The readme part looks fine, so just the tests left :)

I'm not sure why you get this test failure, but your code shouldn't cause existing tests to fail.
To add a test for your feature you could follow the approach [here](https://github.com/thlorenz/doctoc/blob/master/test/transform-title.js). It's also passing a flag to modify the outcome ( in this case `--title`).

package.json
```diff
@@ -8,7 +8,8 @@
     "url": "git://github.com/thlorenz/doctoc.git"
   },
   "scripts": {
-    "test": "set -e; for t in test/*.js; do node $t; done"
+    "test": "set -e; for t in test/*.js; do node $t; done",
+    "test:cov": "set -e; tap test/*.js --cov --coverage-report=lcov;"
```
thlorenz: That's not part of this feature. Please remove from now.
You can propose test coverage in another PR and we can think about if we want to pull that.

karfau:
OK, this helped a lot, and you can see what I added.

It even helped me to think about the case where passing file argument to `doctoc`, which is actually being very explicit by itself. See the updated README related to that.

Please have a look if the tests are enough.
(I can imagine more test cases, especially when combining certain flags. But currently the only way to test the code changes I made in `doctoc.js` it by adding more "end-to end" tests like the ones for `stdout` flag.)

package.json
karfau: Should I rebase/squash to make those changes disappear or just add a new commit to undo those changes?

README.md
```diff
@@ -52,7 +51,20 @@ specified](#using-doctoc-to-generate-links-compatible-with-other-sites).
...
+Doctoc locates the TOC by the `<!-- START doctoc -->` and `<!-- END doctoc -->` comments, 
+so you can also move a generated TOC to any other portion of your document and it will be updated there.
+
+When passing a directory to doctoc you can tell it to **only update** files that **already** have a TOC 
+(including those that [specify location of toc](#specifying-location-of-toc)),
+by adding the `--explicit` flag:
+
+    doctoc . --explicit
+
+When being ***explicit*** by passing one ore more file arguments to doctoc,
```
thlorenz: Not a huge fan of that _feature_.
If I pass you only files then I still expect the ones to be affected that already have a TOC since I passed the explicit flag.
You never know where those filenames come from, i.e. could be from a `*.md` pattern expanded in another script.

So please **no magic** ;)

thlorenz:
Tests look good, please fix the minor nits I pointed out (mainly remove magic and - for now - coverage), then squash the commits and we'll get some more peeps to have a look to get this merged.

Thanks for your work so far.

karfau:
OK, ready for more feedback.

thlorenz:
Hey, haven't dropped the ball on this one, just super busy (traveling and such).
I'll have a look once things quiet down (might be a week).

However if others collabs wanna add some feedback in the meantime, that'd be great :)

karfau:
Hey @thlorenz , any news on this one?

Would love to be able to update to a version containing these changes to simplify our scripts.

Best Christian

karfau:
@thlorenz Is there still a chance that this will land?
If there is something I can do, let me know.

Best Christian

karfau:
another ping after another half a year:
@thlorenz any chance this could be merged and released?

mkay581:
would love for this to be merged. Any update?

mkay581:
I stumbled upon this package and actually found it quite puzzling that this wasn't the default behavior 😕

mclarkson:
I have installed from https://github.com/karfau/doctoc. Thankfully it does not conflict with this as the fork installs as `doctoc-explicit`. I would prefer to use the original doctoc (here!) but the forked version makes my workflow much easier. No pipes just:

```
cd blog/dir
doctoc-explicit .
```
I don't want to accidentaly put TOCs on all of my files, which I did a few months back on a large documentation repo. Not such a big deal but I had edited a bunch of documents, ran doctoc, and then had to clean up without messing up my wanted changes.

It would be great if this was the default behaviour.

AndrewSouthpaw:
Looks like this one has been superseded by https://github.com/thlorenz/doctoc/pull/181, I didn't realize there was an older PR available. 😅 

If someone wants to pull the tests in this PR and cut a new one to expand test coverage, that'd be grand. 🙏 

Closing for now.

karfau:
Nice to see something happening again in the repo. I already gave up watching it...

And huge congrats for merging that feature!

As you might have read from the comments, I made a fork (that is outdated now) which used this PR and even made this the default behavior.

Please let me know if there is any interest to add this to this repo.
If so I'm willing to invest time into adding the tests and filing the PR to make it the default.
Otherwise I'll leave it be, since I'm not really using this library anymore.

AndrewSouthpaw:
Hey thanks for following up! I'm trying to pick up some maintenance work on the repo to help out. 😄 

If you can pull in the tests, that'd be great! The functionality has more or less been implemented with a different name (`--update-only` IIRC), but it should basically behave the same...

PR #134, opened 2017-10-05
Add pre-commit metadata to doctoc

asottile:
This enables doctoc to be integrated seamlessly with the following
[pre-commit](http://pre-commit.com) configuration:

```yaml
repos:
-   repo: https://github.com/thlorenz/doctoc
    sha: ...  # substitute a tag here
    hooks:
    -   doctoc
```

If merged, I'll add this to our [growing list of hooks](http://pre-commit.com/hooks.html)!

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,5 @@
+-   id: doctoc
+    name: doctoc
+    language: node
+    entry: doctoc
+    types: [markdown]
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ by github or other sites via a command line flag.
   - [Specifying a custom TOC title](#specifying-a-custom-toc-title)
   - [Specifying a maximum heading level for TOC entries](#specifying-a-maximum-heading-level-for-toc-entries)
   - [Printing to stdout](#printing-to-stdout)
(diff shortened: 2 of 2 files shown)


thlorenz:
Nice .. gonna let this sit for a day or so to give other collabs a chance to object (for whatever reason), but I like it :)

asottile:
@thlorenz any updates on this?  Would love to get this included so I can start using it :D

jez:
I'm still a bit confused: does this affect how people develop in this repo? Or is this mostly so that people can use doctoc as a pre-commit hook on their own projects?

I'm generally against pre-commit hooks that block committing. I don't care for the added slowness of committing, and there are times when I want to commit unfinished changes where the pre-commit hook doesn't pass. 

It sounds like this config is mostly for people who want to use pre-commit hooks elsewhere, as opposed to changing the development workflow for working on doctoc itself. Is this right?

asottile:
This doesn't change anything about doctoc's workflow, just enables others to use it easily.

thlorenz:
Merged, no need to publish to npm right @asottile as it pulls directly from github?

asottile:
@thlorenz correct!  Thanks :D

asottile:
Should show up soon on http://pre-commit.com!  Thanks again: https://github.com/pre-commit/pre-commit.github.io/pull/143

PR #142, opened 2018-03-02
feat: now supports minheaders flag

benjamincharity:
`doctoc —minheaders X .` will only generate a TOC if at least X headings exist

**NOTE:** Functionally this is complete; but I have not yet added tests. I am not very familiar with this testing format and one test already seems to be failing (prior to my change). Any help/direction here would be appreciated! 🙏


PeterDaveHello (commented):
Is it possible that we can get some help from the latest LLM models to add some tests for it?

PR #143, opened 2018-03-05
Handle file skipping internally

zyskowsk:
The current solution for skipping files is to have a system utility such
as `ack` handle the searching for you.  This is not ideal because it
makes the documentation not as clear and ads a dependancies in some
cases.

Here we have the library handle skipping files, but simply not
transforming files that contain the tag

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ by the markdown parser. Doctoc defaults to using the GitHub parser, but other
 [modes can be
 specified](#using-doctoc-to-generate-links-compatible-with-other-sites).
 
+### Ignoring individual files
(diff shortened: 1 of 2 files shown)


wolfcon:
@thlorenz plz accept this.

AndrewSouthpaw (changes requested):
Looks like a step in a great direction, thanks! Please add a test to assert behavior, and I think we'll be all set with this.

wolfcon:
@AndrewSouthpaw I just add a test for this PR. [Check it out](https://github.com/thlorenz/doctoc/pull/206)

AndrewSouthpaw (approved):
LGTM, nice work!

PR #149, opened 2018-05-25
Add to the template warnings for contributors

jamesray1:


jamesray1:
I see that they're automatically generated.

PR #160, opened 2019-02-20
expose API in top level

alexandru-constantin:
creating an access point in the top level of this module to be able to access it via code

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -115,7 +115,4 @@ for (var i = 0; i < argv._.length; i++) {
   console.log('\nEverything is OK.');
 }
 
-module.exports.transform=function(content, mode, maxHeaderLevel, title, notitle, entryPrefix){
-  return transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix);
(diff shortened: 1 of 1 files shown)


doctoc.js
```diff
@@ -114,3 +114,8 @@ for (var i = 0; i < argv._.length; i++) {
 
   console.log('\nEverything is OK.');
 }
+
+module.exports.transform=function(content, mode, maxHeaderLevel, title, notitle, entryPrefix){
+  return transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix);
+};
```
AndrewSouthpaw: You could write this more simply as

```js
module.exports.transform = transform
```

alexandru-constantin: yes i agree. sorry i'm new to nodejs i just needed the API

AndrewSouthpaw: No problem, just go ahead and make that change and I'll merge it in. :)

alexandru-constantin: done!

PR #169, opened 2019-08-13
Don't transform files whose TOC is unchanged

akheron:
Looking at the code, this was always the intention, but there was a small bug in the logic that this PR fixes. Some tests needed to be updated, because they relied on the fact that even unchanged TOC was reported as transformed.

This is helpful e.g. to check in CI that all TOCs have been updated, like this:

    doctoc --stdout . | grep -q "should be updated" && echo "TOCs are out of date!"

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -152,7 +152,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, tit
 
   var wrappedToc =  start + '\n' + toc + '\n' + end;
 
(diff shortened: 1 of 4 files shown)


thlorenz:
@AndrewSouthpaw this looks nice and assuming it doesn't break existing functionality we could merge that in and publish a major version upgrade.
What do you think?

And @akheron sorry for the multi-year delay on my part 😢

PR #178, opened 2020-05-07
pre-commit docs: update sha -> rev

asottile:
Since pre-commit [v1.7.0](https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md#170---2018-03-03), `rev` is the preferred spelling for that field


asottile:
@thlorenz anything else I need here?

AndrewSouthpaw (approved):
LGTM, thanks for the contribution!

PR #181, opened 2020-07-03
feat: add option to keep files with no toc inside untouched

uetchy:
add cli flag to prevent doctoc from adding toc to a file without toc.
this is good for CI or the combination with [`lint-staged`](https://github.com/okonet/lint-staged).
with -p, doctoc will update toc if md has toc sectioon, otherwise doctoc keep it unmodified.

I'm not sure `preserve` is the right name for it. so please change it if you have a better one.

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -62,7 +62,7 @@ function printUsageAndExit(isErr) {
 
   var outputFunc = isErr ? console.error : console.info;
 
-  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] [--all] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
+  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] [--all] [--update-only] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
   outputFunc('\nAvailable modes are:');
   for (var key in modes) {
     outputFunc('  --%s\t%s', key, modes[key]);
--- a/doctoc.js
+++ b/doctoc.js
@@ -20,6 +20,10 @@ function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPref
   if (processAll) {
     console.log('--all flag is enabled. Including headers before the TOC location.')
   }
+
+  if (updateOnly) {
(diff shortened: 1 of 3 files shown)


AndrewSouthpaw (changes requested):
I'd add some additional printout, clarifying that only existing TOC'ed files will be updated. Maybe [here](https://github.com/thlorenz/doctoc/pull/181/files#diff-deb40c96c1eed97349c7c1ad0c1f31c9R43-R44).

Also, I'd suggest a name like `--update-existing` or `--update-only` as that may carry more meaning than "preserve".

Thoughts?

uetchy:
@AndrewSouthpaw Thank you for the suggestion. `--update-only` sounds nice.
Also, I've added the notice and help text for this option.

doctoc.js
```diff
@@ -16,17 +16,21 @@ function cleanPath(path) {
...
 }
 
-function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut, preserve) {
+function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut, updateOnly) {
   if (processAll) {
     console.log('--all flag is enabled. Including headers before the TOC location.')
   }
+
+  if (updateOnly) {
+    console.log('--update-only flag is enabled. Only updating existing TOC.')
```
AndrewSouthpaw, line 25: Nit: "... Only updating files that already have a TOC." I think that reads more clearly.

AndrewSouthpaw (changes requested):
Looks great! Just one minor word tweak.

doctoc.js
```diff
@@ -16,17 +16,21 @@ function cleanPath(path) {
...
 }
 
-function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut) {
+function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut, updateOnly) {
   if (processAll) {
     console.log('--all flag is enabled. Including headers before the TOC location.')
   }
+
+  if (updateOnly) {
+    console.log('--update-only flag is enabled. Only updating existing TOC.')
```
uetchy, line 25:
```suggestion
    console.log('--update-only flag is enabled. Only updating files that already have a TOC.')
```

@AndrewSouthpaw Now sounds concise!

AndrewSouthpaw (approved):
Thanks for your contribution!

PR #183, opened 2020-08-11
add --entire flag

sarisia:
`<!-- START doctoc -->` and `<!-- END doctoc -->` are described as the indicator of TOC location,
but it actually affects TOC generating and it's not documented.

In case we have a line `<!-- START doctoc -->` at the middle of the document,
headings above the line will be omitted from generated TOC:

```markdown
# Heading 1
heading 1

<!-- START doctoc -->
<!-- END doctoc -->

# Heading 2
heading 2
```

```markdown
# Heading 1
heading 1

- [Heading 2](#heading-2)

<!-- END doctoc generated TOC please keep comment here to allow auto update -->

# Heading 2
heading 2
```

I'm not sure this behaviour is intended or not, but, anyway, it's better making this configurable.

This PR adds `--entire` flag to CLI.
With `--entire` flag, the above example will generate the TOC includes headings above
`<!-- START doctoc -->` line:

```markdown
# Heading 1
heading 1

- [Heading 1](#heading-1)
- [Heading 2](#heading-2)

<!-- END doctoc generated TOC please keep comment here to allow auto update -->

# Heading 2
heading 2
```

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -16,13 +16,13 @@ function cleanPath(path) {
   return homeExpanded.replace(/\s/g, '\\ ');
 }
 
-function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, stdOut) {
+function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, entire, stdOut) {
   console.log('\n==================\n');
 
   var transformed = files
     .map(function (x) {
       var content = fs.readFileSync(x.path, 'utf8')
-        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix);
+        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix, entire);
       result.path = x.path;
       return result;
     });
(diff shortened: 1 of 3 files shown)


AndrewSouthpaw (changes requested):
Nice contribution! I didn't even notice that was a thing.

Can you rename the option to `--all` or `--include-headers-before-toc` or something like that? `entire` doesn't make it clear...

I think it'd also be good to print out what it's doing when using the flag. Say "Including headers before the TOC location" or something like that.

With that, I think we'll be good to go.

sarisia:
Thank you for reviewing!

`--all` sounds great! Renamed it.

Also, I've added additional warning when `--all` flag is set, as suggested.
It prints like:
```
$ doctoc --all README.md

DocToccing single file "README.md" for github.com.
--all flag is enabled. Including headers before the TOC location.

==================

"README.md" will be updated
```

By the way, I cannot add commits directly to this PR since the branch `neuenmuller/entire` is used
by other projects, so I have created the new one, #186 and closing this.

PR #186, opened 2020-10-13
add --all flag

sarisia:
Supersedes #183

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -16,13 +16,13 @@ function cleanPath(path) {
   return homeExpanded.replace(/\s/g, '\\ ');
 }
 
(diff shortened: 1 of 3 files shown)


thlorenz:
Thanks for taking care of these @AndrewSouthpaw 
Once you feel like we should publish a new version please LMK (including if it should be major or minor version upgrade).

AndrewSouthpaw:
@thlorenz I think there's one or two other PRs that could be added in that include new flags, at which point a minor release would make sense I think. 😄

AndrewSouthpaw:
@thlorenz I think this is ready for another version bump, I've merged in a few PRs at this point.

PR #216, opened 2021-10-21
Fix several bugs due to line endings not properly handled on Windows platform

### LINKED ISSUE #161: Unusable as 'pre-commit' hook, since it will always fail.
Hi, I would like to point out a potential problem with `pre-commit` usage of `doctoc`. Since `doctoc` always updates files, which contain any headings, *regardless of changes* since the last run, it will *always* modify files and cause the `pre-commit` hook to fail and revert the changes. 

This makes, in my opinion, `doctoc` unusable as a pre-commit hook as is, since it will cause every commit which would update a markdown file fail. The only option you are left with (and please correct me if I'm wrong) is to either skip `pre-commit` on that commit or remove `doctoc` from your `pre-commit-config.yaml`.

### Suggestion

Since `pre-commit` aims more towards hooks which only check but, do not modify any files (such as linters, etc. ) the only way I see `doctoc` being used as a hook, would be if it only updated the files which had their structure changed since the last `doctoc` run or it had a mode in which it would simply run this check and fail, if some files needed updates. (Since both of them require this check to happen, it makes sense to implement it either way)

The way `doctoc` is working right now (and again, this is just *my* opinion and I'm open to being corrected) is not compatible with the direction `pre-commit` is heading. So in my humble opinion, the `pre-commit` integration in the current state is deprecated and it should either be addressed, or marked as suc. Again, I'm open for discussion.

MapleCCC:
In a nut shell, this PR fixes several bugs:

1. Fix the bug that files on Windows always get modified by doctoc even when nothing has changed, and ultimately leading to non-usability of doctoc as pre-commit hook, elaborated on issue #161.

2. Fix the bug that files on Windows, after modified by doctoc, inadvertently get undesirable mixed line endings.

3. Fix the bug that files on Windows with a TOC that has no title, are inferred by doctoc to has a TOC title that contains a single character "\r", which ultimately leads to extra line breaks.

Close #161.

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -2,7 +2,8 @@
 
 'use strict';
 
-var path      =  require('path')
+var os        = require('os')
+  , path      = require('path')
   , fs        =  require('fs')
   , minimist  =  require('minimist')
   , file      =  require('./lib/file')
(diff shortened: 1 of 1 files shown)


doctoc.js
```diff
@@ -16,6 +17,28 @@ function cleanPath(path) {
   return homeExpanded.replace(/\s/g, '\\ ');
 }
 
+function readFile(path, encoding) {
```
AndrewSouthpaw, line 20: I think a name like `xFileWithCRLFHandling` would be more self-documenting.

AndrewSouthpaw (changes requested):
Thanks for the contribution, and sorry for the slow reply!

I worry a little about this being the default behavior always. Perhaps there are Windows users out there who already have this problem solved. Or that as a Windows user perhaps they're still using the regular `\r` in their editors.

So my impression right now is that it isn't safe to turn on for everyone.

If you wanted, you could create a flag to enable the behavior, something like `--useCRLFEndings` or something. Then people could opt into it if that's their desired line endings.

Issue #174, opened 2019-11-13
add bitbucket server mode


dstandish:
with bitbucket server, you have to add anchor tags e.g. `<a name='this-heading'>` for `# This heading`

zhongxiao37:
I made a change in https://github.com/zhongxiao37/doctoc/commit/6d94c0d5543e49807b8c622b511ea6ef3f80aa18 and it worked in my local.

Greg7000:
To work on bitbucket server, TOC must generate anchors for each title. I can't have working links in bitbucket server.

See https://github.com/yzhang-gh/vscode-markdown/issues/703 for more info on this.

PR #218, opened 2022-01-18
resolve issue #174

zhongxiao37:


AndrewSouthpaw (changes requested):
Thanks for the contribution, and sorry for the slow reply! 😅 

This seems great. Please include some tests to validate its behavior, and link to the appropriate documentation for Bitbucket headers. No need to link to the issue.

PeterDaveHello:
@zhongxiao37 would you like to add some tests for it?

PR #225, opened 2022-04-02
fix: upgrade deps

hongaar:
This updates all dependencies to their latest version.

Specifically, this resolves an issue with a vulnerable version of `trim@0.0.1` in the current dependency tree.
```
$ npm ls trim
doctoc@2.1.0 /workspaces/doctoc
└─┬ @textlint/markdown-to-ast@6.1.7
  └─┬ remark-parse@5.0.0
    └── trim@0.0.1 
```

More details here: https://github.com/advisories/GHSA-w5p7-h5w8-2hfq

I had to slightly tweak one test to account for a change in the markdown parser (heading not recognized when part of a previous paragraph).


AndrewSouthpaw:
Oof, sorry for the slow response here. This looks good to me, thanks.

Issue #173, opened 2019-10-31
Add a Dockerfile to containerise the tool


pavelexpertov:
I really like the tool, but my colleagues tend to not have npm/yarn/nvm installed since they aren't JavaScript developers.

Thus, I was thinking to make a pull requests with a simple Dockerfile and it will be up to you to publish an image to registry or not.

pavelexpertov:
Hey @thlorenz , I am about to write a Dockerfile for my team, but I am interested to contribute it to the repository for others to use. But I would like to receive confirmation that you would allow me to provide a pull request for this to happen.

thlorenz:
@pavelexpertov do the images that @PeterDaveHello provided help you?

If so @PeterDaveHello would you mind adding that info to the Readme via a PR please?

PeterDaveHello:
Sure I can add it 👍

PeterDaveHello:
PR sent: #226

PR #226, opened 2022-06-26
Add unoficial Docker image info in README.md

PeterDaveHello:
As @thlorenz suggested in https://github.com/thlorenz/doctoc/issues/173#issuecomment-1164547953, add that info in the README

cc #173


thlorenz:
Thanks!

PR #232, opened 2022-08-06
Upgrade `anchor-markdown-header` to fix emoji-titled headers

### LINKED ISSUE #191: TOC headers with emoji create broken relative-links
When I seem to use emojis like the image below:
image

Doctoc generates the following:
image

The only valid relative-link seems to be 'Staging' while the rest needs manual update to the following to work:
image

Wanted to know if I create a PR that fixes this it would be accepted.

kevin-david:
### Add test for emoji-titled headers

This provides an example of the problem described in https://github.com/thlorenz/doctoc/issues/191.

### Add launch.json to enable test debugging

For codespaces/VS Code. This was very helpful in diagnosing https://github.com/thlorenz/doctoc/issues/191.

Based on approach taken in https://stackoverflow.com/a/39607924, though the port stuff wasn't necessary anymore.

### Upgrade anchor-markdown-header to fix emoji-titled headers 

Actually fixes #191

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -329,11 +329,11 @@
       }
     },
     "anchor-markdown-header": {
-      "version": "0.5.7",
-      "resolved": "https://registry.npmjs.org/anchor-markdown-header/-/anchor-markdown-header-0.5.7.tgz",
-      "integrity": "sha1-BFBj125qH5zTJ6V6ASaqD97Dcac=",
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/anchor-markdown-header/-/anchor-markdown-header-0.6.0.tgz",
+      "integrity": "sha512-v7HJMtE1X7wTpNFseRhxsY/pivP4uAJbidVhPT+yhz4i/vV1+qx371IXuV9V7bN6KjFtheLJxqaSm0Y/8neJTA==",
       "requires": {
-        "emoji-regex": "~6.1.0"
+        "emoji-regex": "~10.1.0"
       }
     },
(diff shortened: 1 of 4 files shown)


thlorenz:
Hi great work @kevin-david on reproducing this in a test.
I'd love if you can take a stab at fixing this as well as I currently don't have bandwidth to address it.
Thanks.

kevin-david:
Hey @thlorenz this should be fixed by my previous change in https://github.com/thlorenz/anchor-markdown-header/pull/45 - I'm actually using my private fork at the moment, but I would love to clean that up!

We just need a new release of https://www.npmjs.com/package/anchor-markdown-header so I can consume it here, is that something you can do? Happy to update this PR to non-draft mode once that's done (with this test now passing)

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
thlorenz, line 1: Please remove this file (just keep a local copy) since we cannot have configs for each editor in here :)

thlorenz (changes requested):
Aside from the vscode config file LGTM, so please just remove and we can merge.
I'm making you contrib so you can do this yourself once you fix that nit.

.vscode/launch.json
kevin-david: Sounds good, removed!

thlorenz:
Thanks, is this ready to be published to npm?

kevin-david:
I think so, I've been testing this for a while with my fork and haven't seen any issues!

PR #235, opened 2022-11-22
fix: formatted headers

hongaar:
Really appreciate this package!

Running into this problem. Currently, headers like 

```md
# foo _bar_
```

are transformed into this TOC entry:

```md
- [foo _bar_](#foo-_bar_)
```

But the actual anchor generated by GitHub is `foo-bar` and the TOC links break. [Here's a sample](https://github.com/hongaar/moker)

~I've added a test to demonstrate this, not quite sure where to start fixing this. I ended up in the [`extractText`](https://github.com/thlorenz/doctoc/blob/bb8296004340eb186406764d7ed1b6edb74fa12a/lib/transform.js#L36) function, however now I'm thinking maybe this should be fixed in https://github.com/thlorenz/anchor-markdown-header?~

~Any help is appreciated.~

**EDIT** This is now addressed with https://github.com/thlorenz/anchor-markdown-header/pull/48. If that lands I will update this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -329,11 +329,12 @@
       }
     },
     "anchor-markdown-header": {
-      "version": "0.6.0",
-      "resolved": "https://registry.npmjs.org/anchor-markdown-header/-/anchor-markdown-header-0.6.0.tgz",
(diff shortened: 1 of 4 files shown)


thlorenz:
Cool, we'll discuss details in the anchor markdown header PR since here we'll just have to bump the version once we figured it out.

flexibrah (approved):

PR #236, opened 2022-11-23
chore: update node lts versions in node.js.yml workflow

hongaar:
Update node versions in default workflow to current LTS releases.

Also including the current release could be added as well with something a little more verbose like:

```yaml
    strategy:
      matrix:
        include:
          - node-version: 14
            lts: true
          - node-version: 16
            lts: true
          - node-version: 18
            lts: false
          - node-version: 19
            lts: false
    continue-on-error: ${{ ! matrix.lts }}
```

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


thlorenz:
Thanks!

PR #249, opened 2023-10-12
fix: Improve `cleanPath` platform compatability

hyperupcall:
Uses `os.homedir()` since that works better on non-POSIXy platforms.

I know this code uses old conventions and methods, so if NodeJS version compatability is a concern, these are when these functions were added to NodeJS:

- [os.homedir](https://nodejs.org/dist/latest-v20.x/docs/api/os.html#oshomedir) was added in `v2.3.0`
- [path.join](https://nodejs.org/dist/latest-v20.x/docs/api/path.html#pathjoinpaths) was added in `v0.1.16`

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -4,13 +4,14 @@
 
 var path      =  require('path')
   , fs        =  require('fs')
+  , os        =  require('os')
   , minimist  =  require('minimist')
   , file      =  require('./lib/file')
   , transform =  require('./lib/transform')
   , files;
 
 function cleanPath(path) {
-  var homeExpanded = (path.indexOf('~') === 0) ? process.env.HOME + path.substr(1) : path;
+  var homeExpanded = (path.indexOf('~') === 0) ? path.join(os.homedir() + path.substr(1)) : path;
 
   // Escape all spaces
   return homeExpanded.replace(/\s/g, '\\ ');


doctoc.js
```diff
@@ -4,13 +4,14 @@
...
   , fs        =  require('fs')
+  , os        =  require('os')
   , minimist  =  require('minimist')
   , file      =  require('./lib/file')
   , transform =  require('./lib/transform')
   , files;
 
 function cleanPath(path) {
-  var homeExpanded = (path.indexOf('~') === 0) ? process.env.HOME + path.substr(1) : path;
+  var homeExpanded = (path.indexOf('~') === 0) ? path.join(os.homedir() + path.substr(1)) : path;
```
PeterDaveHello, line 14: While using `os.homedir()` indeed improves cross-platform compatibility, I noticed that the introduction of `path.join()` deviates slightly from the existing style. I'm unsure if this stylistic change was intended for this PR.

Additionally, if `path.join()` is used, it should accept multiple arguments (e.g., `path.join(os.homedir(), path.substr(1))`) rather than a single concatenated string.

Let me know your thoughts on this!

hyperupcall: Agreed! Embarrassingly, when I first tested these changes I must have forgotten to quote:

```console
$ ./doctoc.js ~/Documents/...
```

```console
$ ./doctoc.js '~/Documents/...'
```

I re-ran and it works as expected. I do prefer `path.join` over concatenation - I fixed up the code to allow for that (renaming the parameter in the function).

doctoc.js
```diff
@@ -4,13 +4,14 @@
...
+  , os        =  require('os')
   , minimist  =  require('minimist')
   , file      =  require('./lib/file')
   , transform =  require('./lib/transform')
   , files;
 
-function cleanPath(path) {
-  var homeExpanded = (path.indexOf('~') === 0) ? process.env.HOME + path.substr(1) : path;
+function cleanPath(filepath) {
+  var homeExpanded = (filepath.indexOf('~') === 0) ? path.join(os.homedir(), filepath.substr(1)) : filepath;
```
PeterDaveHello, line 14: Is renaming `path` to `filepath` a mistake because it's unrelated to the topic?

hyperupcall: Nope it's intentional; I renamed `path` to `filepath` because the `path` parameter in the `cleanPath` function was shadowing the `path` variable defined at the top of the file (`var path = require('path')`). And this caused a bug, since `path.join` is a function exported from the `path` module - `String.prototype.join()` does not exist. A mistake in my testing prevented me from catching this earlier

As I understand both files and folders can be passed to the `cleanPath` function, maybe you prefer a different rename?

PeterDaveHello: Oh, okay, I thought it was about the rewrite of `process.env.HOME` only, will take a deeper look at it.

Joj501 (approved):
Approved to change

doctoc.js
PeterDaveHello: Hi @hyperupcall,

I rechecked the change, and it looks like the naming change is helpful but out of the platform compatibility scope, so I'll respectfully suggest two solutions:

1. Revert the naming change for now, and once this PR is merged, we can have another pull request to handle it.
2. Keep them in this pull request, but rewrite the subject and description to properly and fully cover the naming issue.

It'll be great to ensure the commit history is clean and clear, not just to help debug and understand the code but also to make sure each change is atomic. So, I'll also encourage you to open an issue for the second change and revise its commit message to mention your thoughts. What do you think?

PeterDaveHello: Just a friendly example of how the second commit message might look like:

> Rename `path` parameter to avoid shadowing `path` module
> 
> The `path` parameter in the `cleanPath` function was shadowing
> the `path` module imported at the top of the file. This caused
> issues when calling `path.join()`, as it was mistakenly treated
> as a string method. Renaming the parameter to `filepath` resolves
> this issue and improves code clarity.

Or a shorter version could be:

> Rename `path` to avoid conflict with `path` module
> 
> Renaming the `path` parameter to `filepath` prevents shadowing
> the imported `path` module and fixes related issues.

I hope this could be helpful :smile:

PR #254, opened 2023-12-24
Share instructions for usage with unofficial doctoc Docker container `docker-doctoc`.

tapyu:
I just added a workflow example which can help people to use doctoc with Docker containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,4 +155,41 @@ TOC stays up-to-date.
 
 ### Docker image
 
-There's an unofficial Docker image project for doctoc, if you'd like to use doctoc via Docker or other container based CI/CD pipeline, you can take a look at [PeterDaveHello/docker-doctoc](https://github.com/PeterDaveHello/docker-doctoc)
+There's an unofficial Docker image project for doctoc, if you'd like to use doctoc via Docker or other container based CI/CD pipeline, you can take a look at [PeterDaveHello/docker-doctoc](https://github.com/PeterDaveHello/docker-doctoc). You can create an GitHub workflow to automatically update you table of contents.
+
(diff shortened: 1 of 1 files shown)


tapyu:
I accidentally deleted the forked repo. That is why I am opening this PR once again :)

AndrewSouthpaw:
Thanks for sharing! Since this is an unofficial container, I don't think it warrants adding to the README. However we can leave your PR here and others can refer to it later through searching.

You might consider editing your PR description to refer to some of the keywords so people might stumble across it more easily. 😄

tapyu:
@AndrewSouthpaw , so... That is a problem for me :grimacing

In order to better organize my repositories, forked repos targeting PRs are prefixed with `to-delete` and are deleted once they are merged or closed. Otherwise, I would have several idle PR-aimed forked repositories. Having an opened PR makes no sense for me.

Moreover, if I simply delete the forked repo, this PR will be lost and you won't be able to refer to these modifications. That being said, I suggest you to decide whether accept or reject this PR because sooner or later I will delete my forked repo, and when I do so, this PR will be automatically lost.

Personally, I see no reasons to avoid something that is "unofficial". What is the matter? Unofficial things are awesome. As long as it is working (and it is indeed working), it should be in your `README.md` file to help anyone who wants to use `doctoc`. If you disagree, no problems, but unfortunately, due to my approach, this PR will be lost and you won't be able to refer to these modifications.

AndrewSouthpaw:
Hi there! Thanks for raising the concerns. I was pretty sure the diff sticks around after the forked repo is deleted, and that does appear to be the case. So it seems fine to leave as-is; people will have access to the documentation if they search for it.

PR #255, opened 2023-12-26
Update README.md

tapyu:
I just added a workflow example which can help people to use doctoc with Docker containers.


AndrewSouthpaw:
Thanks for sharing! Since this is an unofficial container, I don't think it warrants adding to the README. However we can leave your PR here and others can refer to it later through searching.

PR #256, opened 2024-01-29
feat: add mdx support

### LINKED ISSUE #211: Support for .mdx files
Hey all, we'd like to use this tool to generate TOCs in a bunch of [`.mdx`](https://mdxjs.com/) files, though it seems like out-of-the-box it only supports `.md` and `.markdown`. How does adding `.mdx` to the list of valid markdown extensions sound? Or, alternatively, a way to pass in user-defined file extension(s) via some CLI flag? (`--ext=mdx` ?)

In the meantime, as a workaround, we've been using something like `find . -name "*.mdx" | xargs doctoc ...`, though it'd be nice to have it handled & supported via `doctoc` itself.

Thanks!

I-3B:
closes #211

[This line](https://github.com/thlorenz/doctoc/compare/master...I-3B:mdx-support?expand=1#diff-387904f6b07263914bc1b0cb291a27832cc6963e01b5d44b75460771fc25ee8fR5) may causes problem if some users where only using doctoc for `.md` files, I didn't  check for it, should I?

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -16,7 +16,7 @@ function cleanPath(path) {
   return homeExpanded.replace(/\s/g, '\\ ');
 }
 
-function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut, updateOnly) {
+function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, stdOut, updateOnly, syntax) {
   if (processAll) {
     console.log('--all flag is enabled. Including headers before the TOC location.')
   }
@@ -24,13 +24,13 @@ function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPref
   if (updateOnly) {
     console.log('--update-only flag is enabled. Only updating files that already have a TOC.')
   }
-  
+
   console.log('\n==================\n');
 
   var transformed = files
     .map(function (x) {
       var content = fs.readFileSync(x.path, 'utf8')
-        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, updateOnly);
+        , result = transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, updateOnly, syntax);
       result.path = x.path;
       return result;
     });
@@ -48,7 +48,7 @@ function transformAndSave(files, mode, maxHeaderLevel, title, notitle, entryPref
     console.log('"%s" is up to date', x.path);
   });
 
-  changed.forEach(function (x) { 
+  changed.forEach(function (x) {
     if (stdOut) {
       console.log('==================\n\n"%s" should be updated', x.path)
     } else {
@@ -62,7 +62,7 @@ function printUsageAndExit(isErr) {
 
   var outputFunc = isErr ? console.error : console.info;
 
-  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] [--all] [--update-only] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
+  outputFunc('Usage: doctoc [mode] [--entryprefix prefix] [--notitle | --title title] [--maxlevel level] [--all] [--update-only] [--syntax (md|mdx)] <path> (where path is some path to a directory (e.g., .) or a file (e.g., README.md))');
   outputFunc('\nAvailable modes are:');
   for (var key in modes) {
     outputFunc('  --%s\t%s', key, modes[key]);
@@ -84,7 +84,7 @@ var mode = modes['github'];
 
 var argv = minimist(process.argv.slice(2)
     , { boolean: [ 'h', 'help', 'T', 'notitle', 's', 'stdout', 'all' , 'u', 'update-only'].concat(Object.keys(modes))
-    , string: [ 'title', 't', 'maxlevel', 'm', 'entryprefix' ]
+    , string: [ 'title', 't', 'maxlevel', 'm', 'entryprefix', 'syntax' ]
     , unknown: function(a) { return (a[0] == '-' ? (console.error('Unknown option(s): ' + a), printUsageAndExit(true)) : true); }
     });
 
@@ -104,7 +104,7 @@ var entryPrefix = argv.entryprefix || '-';
 var processAll = argv.all;
 var stdOut = argv.s || argv.stdout
 var updateOnly = argv.u || argv['update-only']
-
+var syntax = argv['syntax'] || "md"
 var maxHeaderLevel = argv.m || argv.maxlevel;
 if (maxHeaderLevel && isNaN(maxHeaderLevel) || maxHeaderLevel < 0) { console.error('Max. heading level specified is not a positive number: ' + maxHeaderLevel), printUsageAndExit(true); }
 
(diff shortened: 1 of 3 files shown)


I-3B:
@AndrewSouthpaw would you please review this?

chris-dura:
> [This line](https://github.com/thlorenz/doctoc/compare/master...I-3B:mdx-support?expand=1#diff-387904f6b07263914bc1b0cb291a27832cc6963e01b5d44b75460771fc25ee8fR5) may causes problem if some users where only using doctoc for `.md` files, I didn't check for it, should I?

So, if I understand, this means that users would all of a sudden start getting TOCs in their `.mdx` files, which they weren't before... I could see how this would be considered a "breaking change" at worst, and annoying at best to users 😅.

I don't think it's worth forcing a `MAJOR` version bump, so to keep it backwards compatible, you may want to make the MDX support "opt-in", so the `doctoc` command does the same thing by default, and you have to explicitly enable `mdx` support?

I-3B:
> > [This line](https://github.com/thlorenz/doctoc/compare/master...I-3B:mdx-support?expand=1#diff-387904f6b07263914bc1b0cb291a27832cc6963e01b5d44b75460771fc25ee8fR5) may causes problem if some users where only using doctoc for `.md` files, I didn't check for it, should I?
> 
> So, if I understand, this means that users would all of a sudden start getting TOCs in their `.mdx` files, which they weren't before... I could see how this would be considered a "breaking change" at worst, and annoying at best to users 😅.
> 
> I don't think it's worth forcing a `MAJOR` version bump, so to keep it backwards compatible, you may want to make the MDX support "opt-in", so the `doctoc` command does the same thing by default, and you have to explicitly enable `mdx` support?

I can think of two approaches:
- `.mdx` is only matched if `--syntax mdx` is provided, but would also still match `.md`
- the more expected behavior, only match `.mdx` if `--syntax mdx` and only match `.md` and `.markdown` if `--syntax md` (default)

AndrewSouthpaw:
That seems reasonable to me. Not breaking the default behavior is important.

I-3B:
@AndrewSouthpaw hi!

fixed the issue mentioned before here a6ee89ccf7c5f3e0f2bcba4df67bd473979a8b7f, so now mdx files won't get matched unless --syntax mdx is explicitly specified.

AndrewSouthpaw:
Hi there! Sorry it took so long to get back to you. The code looks good, all it needs now is some test coverage and we can ship it! 🥳

I-3B:
Hi~ @AndrewSouthpaw 
added some tests, should cover the new change, would you like me to add more?

lib/transform.js
```diff
@@ -6,18 +6,29 @@ var _             = require('underscore')
...
+var commentEscapedStart = commentTypes.md.escapedStart
 
 function matchesStart(line) {
-  return (/<!-- START doctoc /).test(line);
+  return new RegExp(`${commentEscapedStart} START doctoc `).test(line);
 }
 
 function matchesEnd(line) {
-  return (/<!-- END doctoc /).test(line);
+  return new RegExp(`${commentEscapedStart} END doctoc `).test(line);
```
AndrewSouthpaw, line 31: I don't love this, because `commentEscapedStart` gets re-assigned dynamically later, which isn't obvious. We could make `matchesStart` return a closure, taking the syntax first: `matchesStart = (syntax) => (line) => ...`. But to keep things simple and not block this PR further, I can call this fine for now. Just add a comment above `var commentEscapedStart` calling out this may be reassigned in `transform` below to use a different syntax.

lib/transform.js
```diff
@@ -31,6 +42,19 @@ function isString(y) {
   return typeof y === 'string';
 }
 
+function generateComments(syntax){
+  var commentStart = commentTypes[syntax]?.start || commentTypes.md.start
+  , commentEnd = commentTypes[syntax]?.end || commentTypes.md.end
```
AndrewSouthpaw, line 47: Simplify as `syntax = 'md'` and then call `commentTypes[syntax]`, no need to fallback safely

doctoc.js
```diff
@@ -104,7 +104,7 @@ var entryPrefix = argv.entryprefix || '-';
 var processAll = argv.all;
 var stdOut = argv.s || argv.stdout
 var updateOnly = argv.u || argv['update-only']
-
+var syntax = argv['syntax'] || "md"
```
AndrewSouthpaw, line 107: Run a quick check here to validate correctness, since we don't seem to get safety automatically, something like

```js
if (!SUPPORTED_EXTENSIONS.includes(syntax) throw new Error(...)
```

This will help prevent mistakes from users.

lib/file.js
```diff
@@ -3,57 +3,58 @@ var path  =  require('path')
...
-function separateFilesAndDirs(fileInfos) {
+function separateFilesAndDirs(fileInfos, syntax = "md") {;
   return {
     directories :  _(fileInfos).filter(function (x) {
       return x.isDirectory() && !_(ignoredDirs).include(x.name);
     }),
-    markdownFiles :  _(fileInfos).filter(function (x) { 
-      return x.isFile() && _(markdownExts).include(path.extname(x.name)); 
+    markdownFiles :  _(fileInfos).filter(function (x) {
+      return x.isFile() && _(syntax === 'mdx'? mdxExts : markdownExts).include(path.extname(x.name));
```
AndrewSouthpaw, line 15: This feels brittle, can you do an object action lookup

```js
var extensions = {
  markdownExts: [...],
  mdxExts: [...],
}

// then use extensions[syntax]
```

lib/file.js
```diff
@@ -3,57 +3,58 @@ var path  =  require('path')
  ,  _   =  require('underscore');
 
 var markdownExts = ['.md', '.markdown'];
+var mdxExts = ['.mdx'];
 var ignoredDirs  = ['.', '..', '.git', 'node_modules'];
 
-function separateFilesAndDirs(fileInfos) {
+function separateFilesAndDirs(fileInfos, syntax = "md") {;
```
AndrewSouthpaw, line 9: Can you validate the default value is actually needed for all places you're doing `syntax = "md"`? If we're always passing in a `syntax` variable, the default is not-DRY noise.

test/transform.js
```diff
@@ -470,3 +470,113 @@ check(
...
+                  , undefined
+                  , undefined
+                  , undefined
+                  , undefined
+                  , undefined
+                  , 'md'
+              )
+
+    var lines = res.wrappedToc.split('\n')
+    var commentLines = lines.slice(0,2).concat(lines[lines.length-1])
```
AndrewSouthpaw, line 551: Pull out as a shared test helper, I think it's fine to leave it local to this file.

AndrewSouthpaw (commented):
LGTM! Sorry for a few more comments, I did one more thorough review, and just a few tweaks would make this code cleaner. Thanks for the additional tests. 😄 

@thlorenz would you want to bump and publish the new version number, or should I take care of that? This seems like a minor version bump.

lib/transform.js
I-3B: something like this? 24a6c96410acf3bf88ac51d33179ae1dabfc1780

I-3B:
Hi @AndrewSouthpaw !
thanks for the thorough review, can you please go through the new commits to check if they cover the review?

lib/transform.js
AndrewSouthpaw: Perf!

AndrewSouthpaw:
Perfect! One quick lint cleanup and this is good to go. Thanks @I-3B.

thompson-tomo:
@I-3B & @AndrewSouthpaw i have fixed the merge conflict issues in #304 so this should be good to merge.

Issue #140, opened 2018-02-22
Feature request: specify minimum header level


sinedied:
A missing option that would be useful in my use cases would be to be able to specify a minimum heading level, so that we can ignore level 1 headers for example.

This would compliment the option that sets the maximum header depth also.

jamesray1:
See also https://github.com/thlorenz/doctoc/issues/153.

domenic:
What's curious is that DocToc's own README uses this pattern. But I haven't been able to figure out how to reproduce it, besides manually editing the generated output.

thlorenz:
What I do is to just move the generated TOC below the first header and any future runs of doctoc will automatically ignore that header.
However if someone wants to stab at making this work via a flag I'm totally open to it.
No bandwidth myself though, so hoping for contributions similar to [max-header-level](https://github.com/thlorenz/doctoc#specifying-a-maximum-heading-level-for-toc-entries).

Ju

thompson-tomo:
Have finished the implementation of min level in #368 with the edge case being handled by treating it as sections & not allowing multiple sections.

PR #283, opened 2025-12-02
User definable Min heading

### LINKED ISSUE #278: Skip h1 from being included in toc
I want to be able to exclude h1 headings from my table of contents.

This option should be able to be controlled via the following command/cli arguments:
- H
- noh1

Which if provided is true (should be skipped) otherwise false and it is included.

The solution should consider both h1 & # as being the same thing.

An example of the expected behaviour would be

```
# header

## sub-header
```

Should result in a toc of 1 line "header" aka

```
**Table of contents**
- sub-header

# header

## sub-header

```

thompson-tomo:
Progresses #140
Closes #278 

This adds new functionality which enables the min heading level to be specified. This functionality is particularly useful when you are using h1 for page headers etc.

Have created this as an early pr to get feedback on cli options and what else needs to be done.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -108,11 +108,12 @@ function determineTitle(title, notitle, lines, info) {
   return info.hasStart ? lines[info.startIdx + 2] : defaultTitle;
 }
 
-exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel = 1, title, notitle, entryPrefix, processAll, updateOnly) {
+exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel, title, notitle, entryPrefix, processAll, updateOnly) {
   if (content.indexOf(skipTag) !== -1) return { transformed: false };
 
   mode = mode || 'github.com';
   entryPrefix = entryPrefix || '-';
+  minHeaderLevel = minHeaderLevel || 1;
 
   // only limit *HTML* headings by default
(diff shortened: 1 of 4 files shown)


lib/transform.js
```diff
@@ -108,7 +108,7 @@ function determineTitle(title, notitle, lines, info) {
   return info.hasStart ? lines[info.startIdx + 2] : defaultTitle;
 }
 
-exports = module.exports = function transform(content, mode, maxHeaderLevel, title, notitle, entryPrefix, processAll, updateOnly) {
+exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel = 1, title, notitle, entryPrefix, processAll, updateOnly) {
```
AndrewSouthpaw: I'm not sure this matters, but it seems like this old codebase doesn't use default params. I'm thinking let's avoid them to not run afoul with compatibility questions.

test/transform.js
```diff
@@ -188,6 +189,53 @@ check(
...
+  , 3
+  , 2
+)
+
+check(
+    [ '# H1h'
+    , '## H2h'
+    , '### H3h'
+    , ''
+    , 'Max. level test - hashed'
```
AndrewSouthpaw: Max and min level test?

thompson-tomo:
No worries @AndrewSouthpaw the requested changes have been made.

AndrewSouthpaw (approved):

Issue #252, opened 2023-11-23
Killer


caledrummond90:
from datetime import datetime, timedelta, timezone
from pathlib import Path

from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.asymmetric import rsa
from cryptography.hazmat.primitives.hashes import SHA256
from cryptography.hazmat.primitives.serialization import (
    Encoding,
    NoEncryption,
    PrivateFormat,
)
from cryptography.x509 import (
    CertificateBuilder,
    DNSName,
    Name,
    NameAttribute,
    SubjectAlternativeName,
    random_serial_number,
)
from cryptography.x509.oid import NameOID

# https://cryptography.io/en/latest/x509/tutorial/#creating-a-self-signed-certificate
def generate_keys():
    folder = Path(__file__).parent

    key = rsa.generate_private_key(
        public_exponent=65537,
        key_size=2048,
        backend=default_backend(),
    )
    (folder / "localhost.key").write_bytes(
        key.private_bytes(
            encoding=Encoding.PEM,
            format=PrivateFormat.TraditionalOpenSSL,
            encryption_algorithm=NoEncryption(),
        ),
    )

    subject = issuer = Name(
        [
            NameAttribute(NameOID.COUNTRY_NAME, "IE"),
            NameAttribute(NameOID.ORGANIZATION_NAME, "Scrapy"),
            NameAttribute(NameOID.COMMON_NAME, "localhost"),
        ]
    )
    cert = (
        CertificateBuilder()
        .subject_name(subject)
        .issuer_name(issuer)
        .public_key(key.public_key())
        .serial_number(random_serial_number())
        .not_valid_before(datetime.now(tz=timezone.utc))
        .not_valid_after(datetime.now(tz=timezone.utc) + timedelta(days=10))
        .add_extension(
            SubjectAlternativeName([DNSName("localhost")]),
            critical=False,
        )
        .sign(key, SHA256(), default_backend())
    )
    (folder / "localhost.crt").write_bytes(cert.public_bytes(Encoding.PEM))

Issue #247, opened 2023-07-27
Sent 


AbulKhayer2244:
bitcoin:1Nze2CG4NoAoT8YKBCnmhj8KWk5fxpQq6X![myaddr.png](https://github.com/thlorenz/doctoc/assets/109155430/9f963601-7bcd-4fe5-b96d-f1dc1b47fdaa)

Sent me

Issue #170, opened 2019-09-01
Online DocToc Crashed


jajosheni:
I wanted to use the Online heroku version of [DocToc](http://doctoc.herokuapp.com/). And i pasted a url which belongs to a private Repo.

After that the application crashed and the website shows an error.
Probably the repo status is unchecked and it is assumed that every repo is public.

I solved my problem by installing doctoc by npm but having an online URL is handy.
Awesome work by the way.

PR #298, opened 2025-12-10
Cleanup

thompson-tomo:
This cleanup old named link in readme which is no longer relevant. It also closes invalid/junk issues.

Closes: #170 
Closes: #245 
Closes: #246 
Closes: #247 
Closes: #252 
Closes: #272


AndrewSouthpaw:
Thanks!

PR #312, opened 2026-02-21
feat: Add configurable padding before title #147

### LINKED ISSUE #147: Anticipate and counter rogue blank line between header comment lines (introduced by Prettier)
As you know, [Prettier](https://prettier.io/) cleans `.md` files. One of the actions it will make it will add a an **empty line** between each HTML comment line. Which, in our case, results in:

```
...

...
```

Now, when doctoc is ran next time, it "doesn't see" the `<!-- DON'T EDIT THIS SECTION` line and it erroneously adds another one. We end up with:

```
## Table of Contents

```

Each prettier+doctoc cycle will add one more: line break +`<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->` row.

Now, I [achieved](https://github.com/prettier/prettier/issues/4235) that Prettier would allow comment blocks, ie. `<!-- prettier-ignore-start -->` and `<!-- prettier-ignore-end -->` to prevent this from happening.

However, we should take steps to make `doctoc` more intelligent too and make it tolerate the blank lines between `<!-- START doctoc` and `<!-- DON'T EDIT THIS SECTION`. `doctoc` should not add duplicate line because of a single line break.

Do you think it's technically achievable?

thompson-tomo:
Closes #147

A user can now configure if padding is added before the title of a toc.

This is designed to enable better co-existance with tools such as prettier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,15 @@ By default,
 
 ### Pad table of contents title
 
-Use the `--padtitle` option to add a padding line above the TOC which ensures formatters such as prettier will pass; e.g., `doctoc --padtitle true .`
+Use the `--toctitlepaddingbefore` option to add padding line/s above the TOC which ensures formatters such as prettier will pass; e.g., `doctoc --toctitlepaddingbefore 1 .`
+
+NOTE: Currently it is only supported to add one line before the title.
 
 By default,
 
 - no padding is added above the table of contents title
+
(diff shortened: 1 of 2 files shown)


thompson-tomo:
Agree that we should tackle that when we plan anew major.

Issue #135, opened 2017-11-01
Different headers with punctuation resolve to the same anchor


rteshnizi:
If I have a header C++ and a header C#, they will both resolve to `#c` anchor in TOC. In Github/Gitlab one would be recognized as `#c` and the other `#c-1`

inventhouse:
I just ran into this; I have a notes doc with sections for different projects/topics, many of which have a subsection called "To Do".  One in the middle, I inadvertently titled "To Do:", that one was parsed into a link to the first to-do section and all the to-do links after it were off-by-one compared with the GitHub-rendered anchors.

PR #327, opened 2026-03-13
fix: Handle headings where only symbols differs

thompson-tomo:
Closes: #135

Prior to this change the test would generate:

```
     "- [foo _bar_](#foo-bar)",    
     "- [foo _bar_](#foo-bar-1)",   
     "- [foo **bar**](#foo-bar)",  
     "- [foo ~bar~](#foo-bar)",    
     "- [_foo bar_](#foo-bar)", 
```

And after the change we end up with

```
     "- [foo _bar_](#foo-bar)",    
     "- [foo _bar_](#foo-bar-1)",  
     "- [foo **bar**](#foo-bar-2)",
     "- [foo ~bar~](#foo-bar-3)",  
     "- [_foo bar_](#foo-bar-4)",  
```

Which is what is expected


test/transform.js
```diff
@@ -764,3 +764,22 @@ test('should use {/* */} comments if syntax=mdx', function (t) {
     t.same(commentLines.every((line) => line.startsWith('{/*') && line.endsWith('*/}')), true)
     t.end()
 })
+
+test('\nduplicate titles but but different symbols', function (t) {
```
AndrewSouthpaw: nit

```suggestion
test('\nduplicate titles but with different symbols', function (t) {
```

Issue #279, opened 2025-10-26
Increased compatability when front matter is used


thompson-tomo:
I want the processor to automatically exclude front matter content and instead only process the actual content.

What this would look like is if the first line is --- it should locate the next --- and start from there.

An example of the expected behaviour would be

```
---
# comment
---

# header

## sub-header
```

Should result in a output with 2 lines in the toc aka

```
---
# comment
---

**Table of contents**
- header
  - sub-header

# header

## sub-header

```

See markdown spec. https://www.markdownlang.com/advanced/frontmatter.html

thompson-tomo:
With latest version this is not an issue in update-only mode but only when adding it to all docs. As such should be addressed along with #153

Issue #258, opened 2024-04-17
Extremely slow with 700KB file


drjasonharrison:
``It took me a while to figure out the cause of my pre-commit hooks getting "stuck" on the doctoc (v2.2.0) step but finally realized that doctoc appeared to be processing a 7000KB .md file that wasn't staged for the git commit. The file contained only three heading entries but the last 99% of the file was log lines from a multi hour run of a program.

Unfortunately, doctoc is still taking more than 25 minutes to process four .md files, because one was more than 700KB. Memory usage was also high at 750MB. 

The command according to ps:

```
/home/jason/.cache/pre-commit/repold2ulagt/node_env-16.14.2/bin/node /home/jason/.cache/pre-commit/repold2ulagt/node_env-16.14.2/bin/doctoc file1.md file2.md file3.md file4.md
```

After deleting a bunch of lines in that file, it's now only 6000 lines (from 5x lines) and now 188KB. And doctoc updates the TOC with a few seconds.

thompson-tomo:
@drjasonharrison in the most recent version, I have focused on optimising the execution path. The benefit of this is performance benefits as seen in #351

Issue #153, opened 2018-08-22
[Feature request, help wanted] Add doctoc by default below the title if it's at the top of the page


jamesray1:
Many docs will have a title in the first line of the file. IMO it would be good if doctoc checks to see if this exists, and by default inserts a ToC after it if it doesn't exist rather than before.

PR #337, opened 2026-04-07
chore: Reduce conversion processes

thompson-tomo:
Progresses #153
Progresses #258
Progresses #279

This works to further improve the code path and reduce un-necessary operations during the file handling process.

Previously we had:
- File content was being read
- content was being split based on EOL
- content was being parsed line by line looking for toc
- file content was transformed to ast to get markdown headers
- file content was transformed to ast to get html headers
- toc string built
- toc split into lines
- toc lines spliced into content lines
- content lines merged back into content

Now we just have

- File content was being read
- file content was transformed to ast
- get toc from ast
- get markdown headers from ast
- get html headers from ast
- toc string built
- content built by merging content to prior old toc, new toc & after old toc.

Hence with this new approach the transformation is reduced & ensures that we only touch the toc and not the entire doc.

This will hopefully improve performance of larger files.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -59,7 +59,7 @@ function getTocMarkers(docNodes, pragmaStyle, syntax) {
     position.start.raw = position.start.raw.replace(" generated TOC please keep comment here to allow auto update", "");
     position.end.raw = position.end.raw.replace(" generated TOC please keep comment here to allow auto update", "");
   }
-  else if (pragmaStyle == "legacy" && position?.existing) {
+  else if ((pragmaStyle == "legacy" || !pragmaStyle) && position?.existing) {
     position.start.raw = marker.start;
   }
   return { marker: marker, positions: [position] };
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -167,11 +167,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
   var tocs = getTocMarkers(docNodes, options?.toc?.pragma?.style, syntax);
 
   
-  if (!(tocs?.positions.length > 0) && updateOnly) {
+  if (!(tocs.positions?.length > 0) && updateOnly) {
     return { transformed: false };
   }
   
-  if (!(tocs?.positions.length > 0)) {
+  if (!(tocs.positions?.length > 0)) {
     var insertPos = 0;
     tocs.positions = [{
       end: {
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -237,7 +237,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
   var transformed = !tocPosition?.existing || content.slice(tocPosition.start.range[0], tocPosition.end.range[1]) != wrappedToc;
   
   if (transformed) {
-    if (tocPosition.start.range[0] > 0) { data.push(content.slice(0, tocPosition.start.range[0])); }
+    if (tocPosition.start.range[0] > 0) { data.push(content.slice(0, tocPosition.start.range[0] - 1)); }
     data.push(wrappedToc);
     if (!tocPosition.existing) { data.push(''); }
     data.push(content.slice(tocPosition.end.range[1] + (tocPosition.existing ? 1 : 0)));
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -25,7 +25,7 @@ function isString(y) {
   return typeof y === 'string';
 }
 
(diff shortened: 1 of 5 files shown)


lib/transform.js
```diff
@@ -26,25 +25,38 @@ function isString(y) {
...
+    start: nodes.find(n => n.type === "start")?.node,
+    end: nodes.find(n => n.type === "end")?.node,
+  };
+  var expected = pragmaStyle == "compact" && original?.start ? {
+    start: original.start.replace(" generated TOC please keep comment here to allow auto update", ""),
+    end: original.end.replace(" generated TOC please keep comment here to allow auto update", ""),
+  } : contentGenerator.pragmaMarkers(syntax, pragmaStyle);
 
-  return { start, end };
+  return { expected: expected, orignals: original };
```
AndrewSouthpaw: typo: `originals`

lib/transform.js
```diff
@@ -26,25 +25,38 @@ function isString(y) {
   return typeof y === 'string';
 }
 
-function generateMarkers(lines, info, pragmaStyle, syntax){
-  var start, end, header;
+function getTocs(docNodes, pragmaStyle, syntax) {
```
AndrewSouthpaw: maybe let's call it `getTocMarkers` to be more explicit

lib/transform.js
AndrewSouthpaw: IMO I'm having trouble following the `expected` vs `originals` semantics. Let's either rename to be more clear, or add some comments to capture how you're thinking about this.

Maybe `expected` -> `newTocPragma`, `originals` -> `existingAstNodes`?

lib/transform.js
```diff
@@ -26,25 +25,38 @@ function isString(y) {
...
-    end = marker.end;
-  }
+  var original = nodes.length < 2 ? null : {
+    start: nodes.find(n => n.type === "start")?.node,
+    end: nodes.find(n => n.type === "end")?.node,
+  };
+  var expected = pragmaStyle == "compact" && original?.start ? {
+    start: original.start.replace(" generated TOC please keep comment here to allow auto update", ""),
+    end: original.end.replace(" generated TOC please keep comment here to allow auto update", ""),
+  } : contentGenerator.pragmaMarkers(syntax, pragmaStyle);
```
AndrewSouthpaw: I discovered a bug where a compact pragma style crashes when an existing TOC is present:

In `getTocs()`, when `pragmaStyle == "compact"` and there's an existing TOC, `.replace()` is called directly on the AST nodes (`original.start` and `original.end`), which are objects — not strings. This throws `TypeError: original.start.replace is not a function`.

The fix is to call `.replace()` on `.raw` instead. I've added a failing test (`compact pragma style with existing toc`) that reproduces this bug. If you think this is valid, go ahead and commit the change.

```suggestion
  var expected = pragmaStyle == "compact" && original?.start ? {
    start: original.start.raw.replace(" generated TOC please keep comment here to allow auto update", ""),
    end: original.end.raw.replace(" generated TOC please keep comment here to allow auto update", ""),
  } : contentGenerator.pragmaMarkers(syntax, pragmaStyle);
```

AndrewSouthpaw (changes requested):

lib/transform.js
thompson-tomo: You are right, I missed updating it when I switched to updating the original nodes.

lib/transform.js
thompson-tomo: Done

lib/transform.js
thompson-tomo: Fair point, I have updated the naming to marker & positions

thompson-tomo:
@AndrewSouthpaw can you let me know if you are happy with the new property naming & if there is any other open points.

test/transform.js
```diff
@@ -795,6 +795,65 @@ test('should use {/* */} comments if syntax=mdx', function (t) {
     t.end()
 })
 
+// Bug: compact pragma style with existing toc crashes because .replace() is called
+// on an AST node (object) instead of node.raw (string). This is a regression
+// introduced by the refactor — it passes on master.
+test('compact pragma style with existing toc', function (t) {
```
AndrewSouthpaw:
```suggestion
test('compact pragma style with existing toc', function (t) {
```

AndrewSouthpaw (approved):
LGTM! Nice work.

thompson-tomo:
Thanks. The next one to progress would be #338 for which I don't expect any merge conflicts with this. 🤞

PR #338, opened 2026-04-07
chore: Reduce explicit usage of new lines

thompson-tomo:
This uses an array of toc contents which is only combined into a string at the end which reduces the explicit usage of EOL characters and increases readability.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -224,11 +224,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
     tocLines.push('');
     if (options?.toc?.footer?.content) { tocLines.push(options.toc.footer.content); }
 
-    toc = tocLines.join('\n').trimEnd();
+    toc = tocLines.join('\n');
   }
 
   var data = [];
-  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n\n' : '') + tocPosition.end.raw;
+  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n' : '') + tocPosition.end.raw;
   var transformed = !tocPosition?.existing || content.slice(tocPosition.start.range[0], tocPosition.end.range[1]) != wrappedToc;
   
   if (transformed) {
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -224,11 +224,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
     tocLines.push('');
     if (options?.toc?.footer?.content) { tocLines.push(options.toc.footer.content); }
 
-    toc = tocLines.join('\n');
+    toc = tocLines.join('\n').trimEnd();
   }
 
   var data = [];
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
@thompson-tomo Some new merge conflicts on this, unfortunately.

thompson-tomo:
All fixed. @AndrewSouthpaw & would be good to get your thoughts on #339 which is the successor to this.

lib/transform.js
```diff
@@ -198,10 +198,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
   var wrappedToc;
```
AndrewSouthpaw: btw this var is now dead (shadowed lower on line 232) and can be deleted

lib/transform.js
```diff
@@ -198,10 +198,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
 
   if (headers.length >= minTocItems) {
-    if (options?.toc?.header?.content) {
-      toc = options.toc.header.content + '\n';
-    }
-
+    var tocL = [];
```
AndrewSouthpaw: Avoid arbitrary abbreviation where possible; let's call this `tocLines` or `tocParts` or something.

AndrewSouthpaw (changes requested):
Just a couple small nits

thompson-tomo:
@AndrewSouthpaw feedback implemented

lib/transform.js
thompson-tomo: Done.

AndrewSouthpaw:
Checks are failing btw

thompson-tomo:
Sorry about that @AndrewSouthpaw fixed now.

test/transform.js
```diff
@@ -9,36 +9,29 @@ function check(md, anchors, mode, maxHeaderLevel, minHeaderLevel, minTocItems, t
...
-      , endLines = legacy.end.split('\n')
-      , mdLines = md.split('\n');
-
-    var rig = startLines
-      .concat(anchorLines.slice(0, -2))
-      .concat(endLines)
-      .concat('')
-      .concat(mdLines);
+    // build the expected content
+    var pragma = contentGenerator.pragmaMarkers(syntax || 'md');
```
AndrewSouthpaw, line 14: Hmm I'd rather not build the `eol` based on the implementation output. It's not _exactly_ tautological (because of a detail in the test assertion), but it's close. 

If we want to eventually test eol configs for both `\n` and `\r\n` etc., I'd rather create a test matrix or write them out explicitly.

This will also allow us to drop the `eol` being exposed from `transform`, which is currently only used by the test.

Thoughts?

AndrewSouthpaw (changes requested):

test/transform.js
thompson-tomo: I agree it is not great but couldn't think of a better way as i wanted to avoid adding another argument to the method but will go down that path. Please take another look as I have made the change & have also updated #339

thompson-tomo:
@AndrewSouthpaw i think this is good to merge now.

PR #339, opened 2026-04-07
fix: line endings are set based on file

### LINKED ISSUE #161: Unusable as 'pre-commit' hook, since it will always fail.

thompson-tomo:
Closes: #161 
Closes: #216

Tool now detects what line endings have been used in the content and uses that in the generated toc to strive for consistency.

Blocked by: #337 & #338

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -224,11 +224,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
     tocLines.push('');
     if (options?.toc?.footer?.content) { tocLines.push(options.toc.footer.content); }
 
-    toc = tocLines.join('\n').trimEnd();
+    toc = tocLines.join('\n');
   }
 
   var data = [];
-  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n\n' : '') + tocPosition.end.raw;
+  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n' : '') + tocPosition.end.raw;
   var transformed = !tocPosition?.existing || content.slice(tocPosition.start.range[0], tocPosition.end.range[1]) != wrappedToc;
   
   if (transformed) {
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -224,11 +224,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
     tocLines.push('');
     if (options?.toc?.footer?.content) { tocLines.push(options.toc.footer.content); }
 
-    toc = tocLines.join('\n');
+    toc = tocLines.join('\n').trimEnd();
   }
 
   var data = [];
-  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n' : '') + tocPosition.end.raw;
+  var wrappedToc = tocPosition.start.raw + '\n' + (toc != '' ? toc + '\n\n' : '') + tocPosition.end.raw;
   var transformed = !tocPosition?.existing || content.slice(tocPosition.start.range[0], tocPosition.end.range[1]) != wrappedToc;
   
   if (transformed) {
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -172,7 +172,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
     }
 
     if (options?.toc?.header?.content) { tocL.push(options.toc.header.content); }
-    if (padTitle) { tocL.push(''); }
+    if (padTitle && inferredTitle) { tocL.push(''); }
     if (inferredTitle) { tocL.push(inferredTitle); }
     tocL.push('');
 
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -144,30 +144,9 @@ function determineTitle(nodes, tocPosition, title, notitle, syntax) {
   return paragraphs?.[paragraphs.length - 1]?.raw;
 }
 
-function detectLineEnding(content, defaultEol) {
-  var eol = '', newEol = '';
-
-  for (var i = 0; i < content.length; i++) {
-    if (content[i] === '\r') {
-      if (content[i + 1] === '\n') {
-        newEol = '\r\n';
-        i++;
-      }
-      else { newEol = '\r'; }
-    }
-    else if (content[i] === '\n') { newEol = '\n'; }
-    else { continue; }
-    if (eol && eol != newEol) { return defaultEol; }
-    eol = newEol;
-  }
-
-  return eol || defaultEol;
-}
-
 exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel, minTocItems, title, notitle, entryPrefix, processAll, updateOnly, syntax, options) {
   syntax = syntax || "md";
   var eol = '\n';
-  eol = detectLineEnding(content, eol);
   var skipTag = contentGenerator.skipTag(syntax) + eol;
   var index = content.indexOf(skipTag);
   if (index === 0 || (index >= 0 && content[index-1] === eol)) return { transformed: false };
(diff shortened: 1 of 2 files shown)


lib/transform.js
```diff
@@ -117,13 +117,30 @@ function determineTitle(title, notitle, lines, info) {
...
+  const lf = (content.match(/\n/g) || []).length - crlf;
+
+  // Count all \r, then subtract CRLF occurrences
+  const cr = (content.match(/\r/g) || []).length - crlf;
+
+  if (crlf > lf && crlf > cr) return '\r\n';
+  if (lf > crlf && lf > cr)   return '\n';
+  if (cr > crlf && cr > lf)   return '\r';
+
+  return '\n';
```
thompson-tomo: Should this default be user definable? Cc @AndrewSouthpaw

lib/transform.js
```diff
@@ -117,13 +117,30 @@ function determineTitle(title, notitle, lines, info) {
...
+
+  // Count all \n, then subtract CRLF occurrences
+  const lf = (content.match(/\n/g) || []).length - crlf;
+
+  // Count all \r, then subtract CRLF occurrences
+  const cr = (content.match(/\r/g) || []).length - crlf;
+
+  if (crlf > lf && crlf > cr) return '\r\n';
+  if (lf > crlf && lf > cr)   return '\n';
+  if (cr > crlf && cr > lf)   return '\r';
```
thompson-tomo: I am not sure about how we should handle files with mixed endings. Perhaps we should do.

```suggestion
  if (lf + cr === 0 && crlf > 0)   return '\r\n';
  if (crlf + cr === 0 && lf > 0)   return '\n';
  if (crlf + lf === 0 && cr > 0)   return '\r';
```

Thoughts @AndrewSouthpaw

AndrewSouthpaw: Hmmm this could be a major change, as I mentioned [here](https://github.com/thlorenz/doctoc/pull/216#pullrequestreview-860417728).

IMO it'd make sense for this to be behind a flag for v2, and change the default to automatically detect for v3.

For files with mixed line endings, I'm leaning towards throwing, because there's not really a reasonable thing to do. Whatever we choose, it could be the wrong behavior for the user, and quite subtle. We can give a flag for them to choose the override (below).

Printing a warning would be another option, but I'm generally a fan of failing fast, especially given that mixed line endings is probably an indication that something is weird and wrong about their file, and it'd be helpful for them to know about it.

Thoughts?

lib/transform.js
```diff
@@ -117,13 +117,30 @@ function determineTitle(title, notitle, lines, info) {
...
+  const crlf = (content.match(/\r\n/g) || []).length;
+
+  // Count all \n, then subtract CRLF occurrences
+  const lf = (content.match(/\n/g) || []).length - crlf;
+
+  // Count all \r, then subtract CRLF occurrences
+  const cr = (content.match(/\r/g) || []).length - crlf;
+
+  if (crlf > lf && crlf > cr) return '\r\n';
+  if (lf > crlf && lf > cr)   return '\n';
```
AndrewSouthpaw: In the spirit of optimization for large files, we could avoid 3 scans of regex matches like this:

```ts
  function detectLineEnding(content) {
    var crlf = 0, lf = 0, cr = 0;
    for (var i = 0; i < content.length; i++) {
      if (content[i] === '\r') {
        if (content[i + 1] === '\n') { crlf++; i++; }
        else { cr++; }
      } else if (content[i] === '\n') { lf++; }
    }
```

As much as I hate for loops 😅

lib/transform.js
AndrewSouthpaw: That would ease some of what I mention above. Something like --line-ending-default or something, which we can fall back to (and not throw) in cases of mixed line endings.

lib/transform.js
```diff
@@ -174,31 +188,30 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
-      end = options.toc.footer.content + '\n' + end;
-    }
-    
+        }));
+
+    tocL.push('');
+    if (options?.toc?.footer?.content) { tocL.push(options.toc.footer.content); }
+
+    toc = tocL.join('\n');
     wrappedToc = start + '\n' + toc + '\n' + end;
```
AndrewSouthpaw: I'm guessing this isn't yet done b/c it's still in draft form, but lines like these need to be updated, otherwise `\r\n` files still get marked as modified:

```js
test('up-to-date toc in \\r\\n file is not marked as transformed', function (t) {
  var mdCrlf = [
      '<!-- START doctoc generated TOC please keep comment here to allow auto update -->'
    , '<!-- DON\'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->'
    , '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*'
    , ''
    , '- [Header](#header)'
    , ''
    , '<!-- END doctoc generated TOC please keep comment here to allow auto update -->'
    , ''
    , '## Header'
    , 'some content'
    , ''
    ].join('\r\n')

  var resCrlf = transform(mdCrlf)
  t.same(resCrlf.transformed, false, '\\r\\n version should also be up to date')
  t.end()
})
```

thompson-tomo: Yeap this PR will need a rebase once the prior is merged, verification all instances are caught & tests added.

lib/transform.js
thompson-tomo: I agree #216 would be a major change especially as it uses the system default rather than detecting and it also re-writes the entire doc.

I think it would be ok if document is consistent use what is detected, otherwise use configured which has a default of '\n'.

This would mean:

- \n docs already with upto date toc would always use \n
- \r\n docs already with upto date toc would use config option with default \n. Can Result in mixed doc as per now if not configured otherwise.
- \r\n docs already with toc containing normalised line endings via script would use \r\n, changed behaviour. This would be low likelihood but what user wants as it removes need to add in a normalisation process. Effectively a bug fix.
- \r\n doc without toc would use \r\n, changed behaviour. In this case it is a change but avoids creating a mixed document which I see as a positive.
- \n doc without toc would use \n, no changed behaviour.


So we would only get a chance in a few scenarios. Let me know if you are on the same page.

lib/transform.js
thompson-tomo: I agree, byte by byte would be more efficient.

lib/transform.js
thompson-tomo: Thinking about this some more, I feel we can merge this without the configurable default. That way this can be released as part of 2.4.1 due to it purely being a fix. This is because the only time a user would see a change is if they are post processing the toc to normalize the endings and all they would see is their processing would have no effect and for new docs we are avoiding introducing a line ending mismatch.

lib/transform.js
thompson-tomo: Let's push configurable option to next minor.

thompson-tomo:
@AndrewSouthpaw now that #338 is merged this one is good for review.

AndrewSouthpaw (approved):

PR #341, opened 2026-04-15
fix: correctly determine title #340

### LINKED ISSUE #340: Duplicated "DON'T EDIT THIS SECTION" lines in README since v2.4.0
After upgrading to doctoc v2.4.0, the "DON'T EDIT THIS SECTION" header lines are being duplicated every time I run the command. This behavior does not occur in v2.3.0.

**Steps to repro**

README.md:

```md
# H1

- [H2-1](#h2-1)
- [H2-2](#h2-2)
- [H2-3](#h2-3)

<!-- END doctoc generated TOC please keep comment here to allow auto update -->

## H2-1

## H2-2

## H2-3
```

Run the latest doctoc:

```bash
npx doctoc@latest README.md --github
```

Observe that the "DON'T EDIT THIS SECTION" lines are added repeatedly instead of being updated in place.

```diff
diff --git a/README.md b/README.md
index 7ddf7a5..f1c9330 100644
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 - [H2-1](#h2-1)
 - [H2-2](#h2-2)
```

**Expected behavior**

The TOC should be updated without duplicating the warning headers, consistent with the behavior in v2.3.0.

**Additional context**

Version 2.3.0: Works as expected (no duplication).

Version 2.4.0: Causes duplication.

Is this an intended breaking change? Also, is there a new option available in v2.4.0 to maintain the previous behavior?

thompson-tomo:
Closes #340

Correctly Handle noTitle existing in toc when in update mode and the notitle argument is not provided.


potiuk (approved):

README.md
```diff
@@ -48,6 +48,7 @@ by github or other sites via a command line flag.
 Use the `--title` option to specify a (Markdown-formatted) custom TOC title; e.g., `doctoc --title '**Contents**' .` From then on, you can simply run `doctoc <file>` and doctoc will keep the title you specified.
 
 Alternatively, to blank out the title, use the `--notitle` option. This will simply remove the title from the TOC.
+It is recommended to also include this argument when performing an update on an existing toc.
```
AndrewSouthpaw, line 51: Fine for it to be separate, but can you provide an example of what `doctoc --title '...'` produces? That should make this option easier to understand.

AndrewSouthpaw, line 51: I think this line is unclear... "an existing toc with a title" maybe? Not exactly sure the case you're describing.

AndrewSouthpaw (approved):
LGTM. 

I have some minor feedback, but you can address this in a follow-up PR so we can get the bugfix out.

I used Claude to summarize the algorithm because sometimes it's hard to hold it all in my head. Here's the response, in case it's useful for future readers:

---

**The bug**: For a no-title TOC like the one in #340:

```

- [H2-1](#h2-1)
...
<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```

The AST has these children between the markers:
- START Html node at range `[X, Y]`
- DON'T EDIT Html node at range `[A, B]`
- List node at range `[C, D]`

`determineTitle` looks for the "title" by finding nodes between the start marker and the list. The **old** filter was:

```js
x.range[0] > tocPosition.start.range[0]
```

That's "anything that starts after the *beginning* of the START marker." The DON'T EDIT node starts after the START marker's beginning, so it gets picked up as the "title" — and then gets inserted into the TOC output, duplicating it.

**Fix part 1 — range update (line 64)**:

In `getTocMarkers`, when the legacy pragma is rewritten to canonical form, `position.start.raw` gets set to `marker.start` (line 63). `marker.start` is the full two-line block:
```
```

The new line 64 updates the range to match:
```js
position.start.range[1] = position.start.range[0] + marker.start.length;
```

Now `range[1]` points past the end of the DON'T EDIT line.

**Fix part 2 — filter change (line 143)**:

The filter changes from `> tocPosition.start.range[0]` to:
```js
x.range[0] >= tocPosition.start.range[1]
```

This means "anything that starts at or after the *end* of the start marker." Since the start marker's range now encompasses both the START and DON'T EDIT lines, the DON'T EDIT node falls *within* the marker and is excluded from the paragraph search.

For the no-title case, no paragraphs exist between the (expanded) start marker and the list, so `paragraphs` is empty and `return title || ""` returns `""` — no title, no duplication.

**Fix part 3 — header content guard (line 147)**:

```js
if (title && header?.includes(title)) return "";
```

Separate edge case: if `--toc-header-content` was used and the inferred "title" matches it, return empty to avoid a different duplication path.

PR #342, opened 2026-04-15
Add regression test for custom header title bug and additional comments

AndrewSouthpaw:
Adds some comments on how the title algorithm works, plus a regression test guarding against one of the bugs fixed by #340.


AndrewSouthpaw:
cc @thompson-tomo would you mind giving a review?

thompson-tomo (approved):

PR #344, opened 2026-04-17
feat: Support frontmatter in docs

### LINKED ISSUE #279: Increased compatability when front matter is used

thompson-tomo:
Closes #279

Insert toc after front matter content to ensure front matter remains valid.


lib/transform.js
```diff
@@ -182,13 +182,23 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
-  
   if (!(tocs.positions?.length > 0) && updateOnly) {
     return { transformed: false };
   }
+
+  var startPos = 0;
+  if (docNodes.children[0].type == 'Yaml') {
+    startPos = docNodes.children[1].range[0];
+  }
+  else if(docNodes.children[0].type == 'Paragraph' && docNodes.children[0].raw.startsWith('+++') && docNodes.children[0].raw.endsWith('+++')){
```
AndrewSouthpaw: use `===`

AndrewSouthpaw (changes requested):
- Looks like it crashes for an empty file
- Crashes for a file with only the YAML frontmatter and no other content
- Same for JSON
- Probably add a test for re-running on a TOC'ed frontmatter file?
- Test case with `processAll`?
- The startsWith/endsWith check on .raw breaks if the closing delimiter has trailing whitespace, e.g. `"+++\ntitle: test\n+++ \n"`

thompson-tomo:
@AndrewSouthpaw i believe all the additional scenarios are being catered for now with the additional tests as well as robustness improvements

lib/transform.js
```diff
@@ -182,15 +182,37 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
+  }
+  else if(hasBody && docNodes.children[0].type === 'Paragraph' && docNodes.children[0].raw.startsWith(';;;') && docNodes.children[0].raw.trim().endsWith(';;;')) {
+    frontMatter = true;
+  }
+  if (frontMatter && docNodes.children.length > 1) {
+    startPos = docNodes.children[1].range[0];
+  }
+  // force it on to the next line with a line between
+  else if (frontMatter) {
+    startPos = docNodes.children[0].range[1] + 2;
```
AndrewSouthpaw: I think this assumes `\n`, right? That'd break for `\r\n`?

lib/transform.js
```diff
@@ -182,15 +182,37 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
   var hasBody = docNodes.children.length > 0;
+  if (hasBody && docNodes.children[0].type === 'Yaml') {
+    frontMatter = true;
+  }
+  else if(hasBody && docNodes.children[0].type === 'Paragraph' && docNodes.children[0].raw.startsWith('+++') && docNodes.children[0].raw.trim().endsWith('+++')) {
+    frontMatter = true;
+  }
+  else if(hasBody && docNodes.children[0].type === 'Paragraph' && docNodes.children[0].raw.startsWith(';;;') && docNodes.children[0].raw.trim().endsWith(';;;')) {
+    frontMatter = true;
+  }
```
AndrewSouthpaw: maybe cleaner:

```js
  var frontMatterDelimiters = ['+++', ';;;'];
  var firstChild = hasBody && docNodes.children[0];
  if (firstChild?.type === 'Yaml') {
    frontMatter = true;
  } else if (firstChild?.type === 'Paragraph') {
    frontMatter = frontMatterDelimiters.some(d =>
      firstChild.raw.startsWith(d) && firstChild.raw.trim().endsWith(d));
  }
```

AndrewSouthpaw (commented):
two small notes, looks just about ready!

lib/transform.js
thompson-tomo: That is true, but I have reworked it to consider this.

AndrewSouthpaw (changes requested):
Nice! It looks like there's one regression worth addressing here. The frontmatter doc/body split in check() above assumes the
closing delimiter is followed by *exactly one* blank line:

```js
var pos = md.indexOf(tag,3) + 3 + 2 * eol.length;
```

With anything other than one blank line, `pos` is wrong (here, 0 blank lines drags the first char of the next line into the doc segment).

Separately, `md.substring(0, 3)` matches any 3-char prefix in `['---', ';;;', '+++']`, so a horizontal-rule "---" at the top of a doc is misclassified as frontmatter and falls into the same broken path. Both should round-trip once the helper scans for the next non-blank line after a verified closing tag.

A couple sample checks, which both fail right now:

```js
// 0 blank lines after the closing delimiter
check(
  [ '---'
  , 'title: Yaml'
  , '---'
  , '# H'
  , 'body'
  ].join('\n')
, [ '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*\n\n'
  , '- [H](#h)\n\n\n'
  ].join('')
)

// "---" at the top is a horizontal rule, not frontmatter — helper false-positive
check(
  [ '---'
  , ''
  , '# Real Heading'
  , 'body'
  ].join('\n')
, [ '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*\n\n'
  , '- [Real Heading](#real-heading)\n\n\n'
  ].join('')
)
```

I was inclined to let it be a follow-up, as I think it's maybe unlikely for docs to have a leading `---` without it being frontmatter... But it would be quite unfortunate if they did. Let's do one more round and address this as in scope to the PR.

thompson-tomo:
Thanks @AndrewSouthpawfor picking up those cases. I have added those checks and am ensuring they are passing.

lib/transform.js
```diff
@@ -201,14 +202,28 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
+  if (frontMatter && docNodes.range[1] === docNodes.children[0].range[1]) {
+    startPos = startPos + eol.length;
+    content += eol;
   }
   
+  // if there is no empty line between front matter & content add one
+  if (frontMatter && hasBody && docNodes.children[0].loc.end.line + 1 == docNodes.children[1].loc.start.line) {
+    content = content.substring(0, startPos) + eol + content.substring(startPos);
+    startPos = startPos + eol.length;
+  }
```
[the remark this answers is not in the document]
thompson-tomo: Not applicable.

lib/transform.js
```diff
@@ -201,14 +202,28 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
   // force it on to the next line after current content
   else if (frontMatter) {
     startLine = docNodes.children[0].loc.end.line;
+    startPos = docNodes.range[1] + 1;
+  }
+
+  // adds in the empty line between end of content & new toc if missing
+  if (frontMatter && docNodes.range[1] === docNodes.children[0].range[1]) {
+    startPos = startPos + eol.length;
+    content += eol;
```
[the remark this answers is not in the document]
thompson-tomo: This is already catered for.

thompson-tomo:
Good news with https://github.com/textlint/textlint/pull/2015 we can natively detect know if json/toml, hence ci broken for package to be updated.

thompson-tomo:
@AndrewSouthpaw could you please give this another look over.

thompson-tomo:
@AndrewSouthpaw could you take a look as this should now be good to merge along with the couple of pr's I have open.

AndrewSouthpaw (commented):
almost there, I think there's another edge case where a stray `\r\n` can lurk:

```js
test('\ngiven a file with CRLF line endings and yaml frontmatter', function (t) {
  var eol = '\r\n';
  var content = ['---', 'title: Yaml Front Matter', '---', '', '# Heading', '', 'Your regular Markdown content follows...', ''].join(eol);
  var res = transform(content);

  t.same(
      res.data
    , [ '---',
        'title: Yaml Front Matter',
        '---',
        '',
        '<!-- START doctoc generated TOC please keep comment here to allow auto update -->',
        '<!-- DON\'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->',
        '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*',
        '',
        '- [Heading](#heading)',
        '',
        '<!-- END doctoc generated TOC please keep comment here to allow auto update -->',
        '',
        '# Heading',
        '',
        'Your regular Markdown content follows...',
        '' ].join(eol)
    , 'generates correct toc preserving CRLF line endings with frontmatter'
  )

  t.end()
})
```

thompson-tomo:
There you are @AndrewSouthpaw have added the provided test & gotten it passing. Turns out there was 2 scenarios which were still assuming \n with both now fixed.

AndrewSouthpaw (approved):

PR #345, opened 2026-04-17
feat: Add toc based on minLines

### LINKED ISSUE #152: [Feature request, help wanted] Filter doctoc by number of lines in a file
Filter doctoc, by default, to not add a ToC to files that are below 30 lines (20 lines would fit on a page on most screens); scrolling 10 lines is just about as easy as clicking on a link, then add a flag to change this number of lines.

I will circulate this to see if anyone has more time to work on this, and will add it to Pocket in case I can find time.

thompson-tomo:
Closes: #152

This adds the TOC only when the document has the required number of lines excluding both front matter and also any existing toc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ by github or other sites via a command line flag.
   - [Max. heading level](#max-heading-level)
   - [Include all headings](#include-all-headings)
   - [Min. ToC items](#min-toc-items)
+  - [Min. Document Lines](#min-document-lines)
   - [Pad table of contents title](#pad-table-of-contents-title)
   - [Indentation Style](#indentation-style)
   - [TOC Pragma style](#toc-pragma-style)
--- a/README.md
+++ b/README.md
@@ -116,6 +116,24 @@ By default,
 
 - The min items is set to 1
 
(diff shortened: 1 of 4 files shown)


AndrewSouthpaw (commented):
I'm not convinced of the value here. What if the lines are super long and wrap, taking up vertical space? What if they're images?

It's a ok idea I suppose but seems like there was only one person asking for it a long time ago. I don't want to overwhelm the lib with options, you know?

thompson-tomo:
While I would agree there is a number of caveats which should be made clearer I still it as useful especially as we get the solution almost for nothing once the frontMatter pr lands.

For instance I don't see line length as being an issue as we can make clear that we expect lines are wrapped. Images are also not that much of an issue as long as we make clear it is based on number of lines in the markdown file.

AndrewSouthpaw (approved):
Hm alright. Needs to resolve merge conflicts then we're good to go.

thompson-tomo:
Thanks @AndrewSouthpaw the Conflicts have been resolved & the documentation improved.

AndrewSouthpaw (approved):

PR #347, opened 2026-04-17
feat: support header id's in header text

### LINKED ISSUE #343: Support Markdown Heading Id's
We should support reading heading id's from markdown headers as described in https://www.markdownlang.com/extended/heading-ids.html

Key thing is the syntax is {#xyz} and {:xyz}

thompson-tomo:
Closes: #343

This verifies that header id's are being parsed


thompson-tomo:
@AndrewSouthand this one is now ready for review.

AndrewSouthpaw (approved):

PR #351, opened 2026-04-25
test: Large file handling

### LINKED ISSUE #258: Extremely slow with 700KB file

thompson-tomo:
Closes #258 

We now run a test against a file with 45k lines, 2k headings & 2.5mb to see how it behaves with such a file.

This test is completing within a matter of seconds (tap has a 30second limit) hence recent changes are resulting in efficient handling.


test/transform-dryrun.js
```diff
@@ -103,3 +103,16 @@ test('\nshould exit with no error code', function (t) {
       }
     })
 })
+
+test('\nshould print to stdout without error and not take too long', function (t) {
+
+    exec('node doctoc.js test/fixtures/k8s-metrics.md --dryrun', function (error, stdout, stderr) {
```
AndrewSouthpaw: nit let's rename the file to be more obvious, e.g. `very-large-file-for-perf-testing.md`

AndrewSouthpaw (approved):
Nice! Cool to see all your hard work paying off. 😄

test/transform-dryrun.js
thompson-tomo: Done.

thompson-tomo:
Yeap it was surprising that it has made as much of an improvement as it has been.

Rename is done.

AndrewSouthpaw (approved):

PR #352, opened 2026-04-25
chore: refactor recursive search for markdown files

thompson-tomo:
Eliminate the statSync to optimise performance as that is often a costly operation which we do not need to be calling.

Overall optimisation of the get recursive ensuring streamlined operations and reduction in variable usage/allocations.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -21,13 +21,13 @@ function findRec(paths, syntax) {
     for (var fileInfo of fileInfos) {
       var fullPath = path.join(currentPath, fileInfo.name);
 
-      if (entry.isDirectory()) {
+      if (fileInfo.isDirectory()) {
         if (!ignoredDirs.includes(fileInfo.name)) {
           stack.push(fullPath);
         }
       }
 
-      if (entry.isFile()) {
+      if (fileInfo.isFile()) {
         if (allowedExts.includes(path.extname(fileInfo.name))) {
           markdownFiles.push({
             path: fullPath,
--- a/lib/file.js
+++ b/lib/file.js
@@ -8,30 +8,48 @@ var extensions = {
   md: ['.md', '.markdown'],
 };
 
-function findRec(currentPath, syntax) {
-  var fileInfos = fs.readdirSync(currentPath, { withFileTypes: true });
-  var markdownFiles = fileInfos.filter(x =>
-      x.isFile() && (extensions[syntax] || []).includes(path.extname(x.name))
-    )
-    .map(x => ({
-      path: path.join(currentPath, x.name),
-      name: x.name
-    }));
-  var markdownsInSubdirs = fileInfos.filter(x =>
-      x.isDirectory() && !ignoredDirs.includes(x.name)
-    )
-    .map(d => findRec(path.join(currentPath, d.name), syntax));
+function findRec(paths, syntax) {
+  var allowedExts = extensions[syntax] || [];
+  var results = [];
+  var stack = paths;
 
-  if (markdownFiles.length > 0)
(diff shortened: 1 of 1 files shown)


lib/file.js
```diff
@@ -8,53 +8,44 @@ var extensions = {
...
+    var fileInfos = fs.readdirSync(currentPath, { withFileTypes: true });
+
+    for (var fileInfo of fileInfos) {
+      var fullPath = path.join(currentPath, fileInfo.name);
+
+      if (fileInfo.isDirectory() && !ignoredDirs.includes(fileInfo.name)) {
+        stack.push(fullPath);
+      }
+
+      if (fileInfo.isFile() && allowedExts.includes(path.extname(fileInfo.name)) {
```
AndrewSouthpaw: missing a closing `)`, syntax error.

lib/file.js
```diff
@@ -8,53 +8,44 @@ var extensions = {
...
-  }
-
-  function process (fileInfos) {
-    var res = separateFilesAndDirs(fileInfos, syntax);
-    var tgts = res.directories.map(d => d.path);
-
-    if (res.markdownFiles.length > 0)
-      log.debug('\nFound %s in "%s"', res.markdownFiles.map(f => f.name).join(', '), currentPath);
+function findRec(paths, syntax) {
+  var allowedExts = extensions[syntax] || [];
```
AndrewSouthpaw, line 12: let's making a copy of the `paths` to avoid mutating the param, slightly less spooky

lib/file.js
```diff
@@ -8,53 +8,44 @@ var extensions = {
...
+    results.push(...markdownFiles);
   }
 
-  var stats = fs.readdirSync(currentPath).map(getStat),
-    res = process(stats),
-    markdownsInSubdirs = res.subdirs.map((subdir)=> findRec(subdir, syntax)),
-    allMarkdownsHereAndSub = res.markdownFiles.concat(markdownsInSubdirs);
-
-  return allMarkdownsHereAndSub.flat();
+  return results;
```
AndrewSouthpaw, line 44: Worth noting: this appears to reverse the order of the results, but AFAICT it has little impact. Things that might change:

- `--stdout` mode will traverse the files in a different order
- Log lines (`X is up to date`) will change

These seem purely cosmetic though and I'm not concerned unless someone raises an issue with it.

AndrewSouthpaw (changes requested):

lib/file.js
thompson-tomo: So stdout, is already limited to only running on a single file so issue 1 shouldn't happen.

Issue 2 I am also not concerned about being an issue.

lib/file.js
thompson-tomo: Aren't we already making a copy of by adding the param into an array on https://github.com/thlorenz/doctoc/pull/352/files#L50 which is the only way to enter this function?

thompson-tomo:
@AndrewSouthpaw this is ready for another look.

lib/file.js
AndrewSouthpaw: Yes, and technically right now it's fine, but it isn't obvious. Mutating a param is quite often an unexpected behavior. There's only you and me working on this codebase, so it's kinda moot, but someone who didn't know any better and didn't look at the contents of a function might use `findRec` and reasonably think `paths` doesn't get affected. It's a general clean coding practice, and in this case I don't see a compelling argument for a perf win to avoid defensive copying.

thompson-tomo: Ok, I think I have changed it as suggested.

AndrewSouthpaw (approved):

Issue #263, opened 2024-10-08
Support specifying the style of the unordered list markup


brlin-tw:
The markdownlint pre-commit hook errored in my project after running `doctoc` because the style of the unordered list is inconsistent with the others:

![Screenshot of the markup of the generated TOC](https://github.com/user-attachments/assets/01c768b2-8cca-4bdf-acf1-8fb50ca129b1)
![Screenshot of the markup of the other content](https://github.com/user-attachments/assets/a2fd1cd3-5dd6-42d8-b7c0-3c2adf9ed6d3)

```txt
$ pre-commit run markdownlint --files README.md
markdownlint.............................................................Failed
- hook id: markdownlint
- exit code: 1

README.md:20:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
README.md:21:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
README.md:41:1 MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
README.md:44:1 MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
```

It would be great if doctoc supports customizing the style of the unordered list so such inconsistency won't occur, specifically:

* The bullet type used in each level of the unordered list
* The indentation width
* The indentation type: space/tab

Thanks in advance!

thompson-tomo:
Indentation width has been implemented, other 2 points are outstanding.

thompson-tomo:
Bullet type support will be available via #353

thompson-tomo:
Bullet type has been merged and will be in the next release which is not far at all from being released.

thompson-tomo:
New release 2.5.0 has occurred which contains the ability to configure symbols.

Indentation style is implemented in #354 hence will hopefully be in the next release.

PR #353, opened 2026-04-26
feat: Add ability for different symbol per level

thompson-tomo:
Progresses #263

This enables the symbol used in unordered toc to differ on a per level basis.

This further assists with adoption as the tool can now work with whatever list symbol/s is used in the doc

### DIFF
--- a/doctoc.js
+++ b/doctoc.js
@@ -132,7 +132,9 @@ for (var key in modes) {
 
 var title = argv.t || argv.title;
 var notitle = argv.T || argv.notitle;
-var entryPrefix = argv.entryprefix || '-';
+var entryPrefix = argv.entryprefix?.trim().replace(' ', '');
+if (entryPrefix?.endsWith(',') || entryPrefix?.includes(',,')) { log.error('Invalid entry prefix: ' + entryPrefix), printUsageAndExit(true); }
+else if(!entryPrefix) { entryPrefix = '-'; }
 var minTocItems = argv.mintocitems || 1;
 if (minTocItems && (isNaN(minTocItems) || minTocItems <= 0)) { log.error('Min. TOC items specified is not a positive number: ' + minTocItems), printUsageAndExit(true); }
 var processAll = argv.all;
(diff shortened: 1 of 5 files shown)


lib/transform.js
```diff
@@ -274,10 +275,14 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
     tocLines.push('');
+    var symbolQty = itemSymbols.length;
 
     tocLines.push(...allHeaders
         .map(function (x) {
-          return indentation.repeat((x.rank - lowestRank) * indentationWidth) + entryPrefix + ' ' + x.anchor;
+          var level = x.rank - lowestRank;
+          var symbol;
+          symbol = itemSymbols[level % symbolQty].trim();
+          return indentation.repeat(level * indentationWidth) + symbol + ' ' + x.anchor;
```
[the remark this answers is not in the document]
thompson-tomo: Handling added.

thompson-tomo:
@AndrewSouthpaw could you take a look at this one?

doctoc.js
```diff
@@ -132,7 +132,9 @@ for (var key in modes) {
 
 var title = argv.t || argv.title;
 var notitle = argv.T || argv.notitle;
-var entryPrefix = argv.entryprefix || '-';
+var entryPrefix = argv.entryprefix?.trim().replace(' ', '');
```
AndrewSouthpaw: `.replaceAll`, otherwise ` "- , * , +"` will fail.

doctoc.js
```diff
@@ -132,7 +132,9 @@ for (var key in modes) {
...
+if (entryPrefix?.endsWith(',') || entryPrefix?.includes(',,')) { log.error('Invalid entry prefix: ' + entryPrefix), printUsageAndExit(true); }
```
AndrewSouthpaw: `",*,+" ` passes validation but would also break.

AndrewSouthpaw (changes requested):

thompson-tomo:
@AndrewSouthpaw changes have been made, thanks for spotting them.

AndrewSouthpaw (approved):
As a follow-up IMO it'd be nice to cover some of those other validation cases

PR #354, opened 2026-04-26
feat: Configure indenting style #263

### LINKED ISSUE #263: Support specifying the style of the unordered list markup

thompson-tomo:
Closes #263

This allows configuration of the indenting style.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "[test/transform-indentation.js]": {
+        "editor.insertSpaces": false,
+    }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "[test/transform-indentation.js]": {
-        "editor.insertSpaces": false
-    }
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
     "[test/transform-indentation.js]": {
-        "editor.insertSpaces": false,
+        "editor.insertSpaces": false
     }
 }
--- a/README.md
+++ b/README.md
@@ -253,14 +253,14 @@ This option also supports customising each level of the list, which can be done 
 
 ##### Indentation Style
 
-Use the `--toc-items-indentation-style` option to control how TOC items are indented, i.e. `doctoc --toc-items-indentation-style space .`.
+Use the `--toc-items-indentation-style` option to control how TOC items are indented, e.g. `doctoc --toc-items-indentation-style space .`.
 
 The supported options are:
 
 - `space`:  The toc item is indented using spaces. This is the default value.
 - `tab`: The toc item is indented using tabs.
 
-The qty of indenting used is calculated by multiplying the `--toc-items-indentation-width` with the toc level.
+The quantity of indenting is controlled by `--toc-items-indentation-width`.
 
 ##### Indentation Width
 
--- a/README.md
+++ b/README.md
@@ -251,6 +251,17 @@ Use the `--entryprefix` option to configure the symbol used in unordered toc, e.
 
 This option also supports customising each level of the list, which can be done by using a comma separated list of symbols, i.e. `doctoc --entryprefix="-,*,+" .`.
 
+##### Indentation Style
+
+Use the `--items-indentation-style` option to control toc items are indented, i.e. `doctoc --items-indentation-style space .`.
+
+The supported options are:
+
+- `space`:  The toc item is indented using spaces. This is the default value.
+- `tab`: The toc item is indented using tabs.
+
+The qty of indenting used is calculated by multiplying the `toc-items-indentation-width` with the toc level.
+
 ##### Indentation Width
 
 Use the `--toc-items-indentation-width` option to customise the indentation width e.g. `doctoc --toc-items-indentation-width 4 .` will set the width to 4.
--- a/README.md
+++ b/README.md
@@ -253,14 +253,14 @@ This option also supports customising each level of the list, which can be done 
 
 ##### Indentation Style
 
-Use the `--items-indentation-style` option to control toc items are indented, i.e. `doctoc --items-indentation-style space .`.
+Use the `--toc-items-indentation-style` option to control how TOC items are indented, i.e. `doctoc --toc-items-indentation-style space .`.
 
 The supported options are:
 
 - `space`:  The toc item is indented using spaces. This is the default value.
 - `tab`: The toc item is indented using tabs.
 
-The qty of indenting used is calculated by multiplying the `toc-items-indentation-width` with the toc level.
+The qty of indenting used is calculated by multiplying the `--toc-items-indentation-width` with the toc level.
 
 ##### Indentation Width
 
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -178,7 +178,14 @@ function detectLineEnding(content, defaultEol) {
 }
 
 exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel, minTocItems, title, notitle, entryPrefix, processAll, updateOnly, syntax, options) {
+  options ??= {};
+  options.toc ??= {};
+  options.toc.items ??= {};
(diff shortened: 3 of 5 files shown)


thompson-tomo:
@AndrewSouthpaw here is the ability to have the toc indented using tabs.

README.md
```diff
@@ -251,6 +251,17 @@ Use the `--entryprefix` option to configure the symbol used in unordered toc, e.
 
 This option also supports customising each level of the list, which can be done by using a comma separated list of symbols, i.e. `doctoc --entryprefix="-,*,+" .`.
 
+##### Indentation Style
+
+Use the `--toc-items-indentation-style` option to control how TOC items are indented, i.e. `doctoc --toc-items-indentation-style space .`.
```
AndrewSouthpaw: `e.g.` is more appropriate than "in essence" here.

```suggestion
Use the `--toc-items-indentation-style` option to control how TOC items are indented, e.g. `doctoc --toc-items-indentation-style space .`.
```

README.md
```diff
@@ -251,6 +251,17 @@ Use the `--entryprefix` option to configure the symbol used in unordered toc, e.
...
+
+The supported options are:
+
+- `space`:  The toc item is indented using spaces. This is the default value.
+- `tab`: The toc item is indented using tabs.
+
+The qty of indenting used is calculated by multiplying the `--toc-items-indentation-width` with the toc level.
```
AndrewSouthpaw: Avoid arbitrary shortening of words, and let's defer that detail to the `--toc-items-indentation-width` section, yeah?

```suggestion
The quantity of indenting is controlled by `--toc-items-indentation-width`.
```

AndrewSouthpaw (changes requested):

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "[test/transform-indentation.js]": {
```
AndrewSouthpaw: I'm guessing this was added to stop an editor from converting the literal tabs in the new test's expected output back to spaces. This only helps if everyone uses VSC, which isn't a guarantee. Can we try using `\t` escapes instead?

lib/transform.js
```diff
@@ -195,12 +195,16 @@ function detectLineEnding(content, defaultEol) {
 }
 
 exports = module.exports = function transform(content, mode, maxHeaderLevel, minHeaderLevel, minTocItems, title, notitle, entryPrefix, processAll, updateOnly, syntax, options) {
-  options ??= {};
-  options.toc ??= {};
-  options.toc.items ??= {};
+  options = options || {};
```
AndrewSouthpaw: Minor: this swaps `??=` for `||` / `== null`, but the rest of this file (and `doctoc.js`) still uses `??=`, `??`, and `?.` freely. I don't think we need it for Node compatibility.

lib/transform.js
```diff
@@ -195,12 +195,16 @@ function detectLineEnding(content, defaultEol) {
...
+  options.toc = options.toc || {};
+  options.toc.items = options.toc.items || {};
+  options.toc.items.indentation = options.toc.items.indentation || {};
+
+  syntax = syntax || 'md';
+  if (options.toc.items.indentation.style == null) options.toc.items.indentation.style = 'space';
+  if (options.toc.items.indentation.width == null) options.toc.items.indentation.width = (mode === 'bitbucket.org' || mode === 'gitlab.com') ? 4 : 2;
```
AndrewSouthpaw: `width` defaults to 2 (4 on bitbucket/gitlab) and indentation is `indentation.repeat(level * width)`. So a user who passes `--toc-items-indentation-style tab` without also setting `--toc-items-indentation-width 1` gets two tabs per level, which they probably don't want.

test/transform-indentation.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+  });
+
+  t.same(
+    headers.toc,
+    [
+        "**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*",
+        "",
+        "- [Hello, world!](#hello-world)",
+        "	- [Installation](#installation)",
```
AndrewSouthpaw: Try using escape characters:

```js
"\t- [Installation](#installation)",
"\t- [API](#api)",
"\t- [License](#license)",
```

AndrewSouthpaw (changes requested):

lib/transform.js
thompson-tomo: We should either set a min node ie 14 in package.json and fix them or set a min node 16 and use coalescing. Will split it off to not block this.

lib/transform.js
thompson-tomo: Been changed so tab default width is 1.

.vscode/settings.json
thompson-tomo: personally i found it confusing having escape characters in expected outputs when the actually character could be used. Yes it only helps vsc users however not all tools seek to standardise spacing vs tabs. To avoid holding this pr up, i have changed it but i think it makes it harder to verify.

lib/transform.js
AndrewSouthpaw: sounds good! I didn't realize coalescing wasn't in 14.

.vscode/settings.json
AndrewSouthpaw: I hear that, but looking at it from the POV of reviewing, escape characters are otherwise invisible in a PR diff. The escape character makes it more explicit.

lib/transform.js
```diff
@@ -198,9 +198,14 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
   options ??= {};
   options.toc ??= {};
   options.toc.items ??= {};
+  options.toc.items.indentation ??= {};
 
   syntax = syntax || "md";
-  options.toc.items.symbols ??= String(entryPrefix || "-").split(',');
+  options.toc.items.indentation.style ??= 'space';
+  if (options.toc.items.indentation.style === 'tab') options.toc.items.indentation.width ??= 1;
+  options.toc.items.indentation.width ??= (mode === 'bitbucket.org' || mode === 'gitlab.com') ? 4 : 2;
```
AndrewSouthpaw: Hmm this works but IMO is a little confusing by leaning on the `??=` here. Could you do something like

```ts
options.toc.items.indentation.width ??= options.toc.items.indentation.style === 'tab' ? 1 : (
  mode === 'bitbucket.org' || mode === 'gitlab.com' ? 4 : 2
)
```

Then it's more clear there's only one evaluation for it.

```suggestion
  options.toc.items.indentation.width ??= options.toc.items.indentation.style === 'tab' ? 1 : (
    mode === 'bitbucket.org' || mode === 'gitlab.com' ? 4 : 2
  );
```

AndrewSouthpaw (changes requested):

lib/transform.js
thompson-tomo: Done, i had wanted to avoid nested ternary but change is done.

AndrewSouthpaw (approved):

lib/transform.js
AndrewSouthpaw: Maybe I'm weird in that nested ternaries are often easier to read to me. 😅  Thanks!

PR #358, opened 2026-06-07
feat: Add support for ordered list

### LINKED ISSUE #257: Feature request: Option for ToC numeration
Would be great if we could generate not only in a form of unordered lists, but also ordered lists (1, 2, 2.1, 2.2, 3, 3.1, 3.1.1, ...).

thompson-tomo:
Closes: #257 

This adds support for ordered lists and not just unordered.

The style of ordered list can be used are:

- number (default)
- uppercase
- lowercase
- roman

The list handles levels ie 1.1.1

Note you must first select ordered as the list format to enable the style to have any effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ Use the `--toc-list-format` option to control the format of the toc list, i.e. `
 The supported options are:
 
  - `ordered`: the list items are numbered rather than using a symbol with the style controlled via the `--toc-list-style` option.
- - `unordered`: the list items use a symnol. This is the default behaviour.
+ - `unordered`: the list items use a symbol. This is the default behaviour.
 
 ##### Style
 
@@ -277,7 +277,7 @@ The supported options are:
  - `lowercase`: the list items are identified using lowercase letters ie `a.c.a`.
  - `number`: the list items are identified using numbers letters ie `1.3.1`. This is the default behaviour.
  - `roman`: the list items are identified using roman numerals ie `I.III.I`.
- - `uppercase`: the list items are identified using lowercase letters ie `A.C.A`.
+ - `uppercase`: the list items are identified using uppercase letters ie `A.C.A`.
 
 #### TOC Items
 
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ Use the `--toc-list-style` option to control the style of the ordered toc list, 
 The supported options are:
 
  - `lowercase`: the list items are identified using lowercase letters ie `a.c.a`.
- - `number`: the list items are identified using numbers letters ie `1.3.1`. This is the default behaviour.
+ - `number`: the list items are identified using numbers ie `1.3.1`. This is the default behaviour.
  - `roman`: the list items are identified using roman numerals ie `I.III.I`.
  - `uppercase`: the list items are identified using uppercase letters ie `A.C.A`.
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ by github or other sites via a command line flag.
     - [General](#general)
     - [Header](#header)
     - [Title](#title)
-    - [List Options](#list-options)
+    - [List](#list)
     - [TOC Items](#toc-items)
     - [Footer](#footer)
 - [Usage](#usage)
@@ -234,7 +234,7 @@ By default,
 
 In all cases there will be padding present after the title due to the toc items always having padding before the list of items.
 
-#### List Options
+#### List
 
 ##### Min. Length
 
@@ -243,6 +243,26 @@ Use the `--mintocitems` option to specify the minimum items required to be in a 
 By default,
 
 - The min items is set to 1
+
+##### Format
+
+Use the `--toc-list-format` option to control the format of the toc list, i.e. `doctoc --toc-list-format unordered`. 
+
+The supported options are:
+
+ - `ordered`: the list items are numbered rather than using a symbol with the style controlled via the `--toc-list-style` option.
+ - `unordered`: the list items use a symnol. This is the default behaviour.
+
+##### Style
+
+Use the `--toc-list-style` option to control the style of the ordered toc list, i.e. `doctoc --toc-list-style roman`. 
+
+The supported options are:
+
+ - `lowercase`: the list items are identified using lowercase letters ie `a.c.a`.
+ - `number`: the list items are identified using numbers letters ie `1.3.1`. This is the default behaviour.
+ - `roman`: the list items are identified using roman numerals ie `I.III.I`.
+ - `uppercase`: the list items are identified using lowercase letters ie `A.C.A`.
 
 #### TOC Items
 
--- a/doctoc.js
+++ b/doctoc.js
@@ -172,7 +172,6 @@ var ordered = format === 'ordered';
 var style = argv['toc-list-style'];
 if (ordered && style && style != 'number' && style != 'uppercase' && style != 'lowercase' && style != 'roman') { log.error('TOC List style: ' + style + ' is not valid.'), printUsageAndExit(true); }
 else if (!ordered && style) { log.error('TOC List style: ' + style + ' should not be specified for unordered list.'), printUsageAndExit(true); }
-else if (ordered && style) { log.error('TOC List style: ' + style + ' should not be specified for unordered list.'), printUsageAndExit(true); }
 
 var entryPrefix = argv.entryprefix?.trim().replaceAll(' ', '');
 if (entryPrefix?.endsWith(',') || entryPrefix?.startsWith(',') || entryPrefix?.includes(',,')) { log.error('Invalid entry prefix: ' + entryPrefix), printUsageAndExit(true); }
--- a/doctoc.js
+++ b/doctoc.js
@@ -132,9 +132,6 @@ for (var key in modes) {
 
 var title = argv.t || argv.title;
 var notitle = argv.T || argv.notitle;
-var entryPrefix = argv.entryprefix?.trim().replaceAll(' ', '');
-if (entryPrefix?.endsWith(',') || entryPrefix?.startsWith(',') || entryPrefix?.includes(',,')) { log.error('Invalid entry prefix: ' + entryPrefix), printUsageAndExit(true); }
-else if(!entryPrefix) { entryPrefix = '-'; }
 var minTocItems = argv.mintocitems || 1;
 if (minTocItems && (isNaN(minTocItems) || minTocItems <= 0)) { log.error('Min. TOC items specified is not a positive number: ' + minTocItems), printUsageAndExit(true); }
 var processAll = argv.all;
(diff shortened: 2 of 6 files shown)


thompson-tomo:
@AndrewSouthpaw i couldn't think of any other common numbering systems to support.

lib/transform.js
```diff
@@ -213,6 +242,11 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
   if (isNaN(minTocItems) || minTocItems < 1){
     minTocItems = 1;
   }
+  
+  if (!options) { options = { }; }
+  if (!options.toc) { options.toc = { }; }
+  if (!options.toc.list) { options.toc.list = { format: 'unordered' }; }
+  if (!options.toc.items) { options.toc.items = { symbols: entryPrefix.split(',') }; }
```
AndrewSouthpaw: These defaults are mostly dead code. By this point `options`, `options.toc` and `options.toc.items` are already non-null and `items.symbols` is populated by lines above, so only the `options.toc.list` default actually does anything.

Also, this default sets `format` but no `style`, and produces a bug:

```
transform(doc, /* 10x undefined */, { toc: { list: { format: 'ordered' } } })
→ ". [My Module]", "  .. [API]"
```

lib/transform.js
```diff
@@ -316,11 +350,31 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
           var symbol;
-          symbol = options.toc.items.symbols[level % symbolQty].trim();
+          if (options.toc.list.format == 'ordered') {
+            var index = (count[level] || 0) + 1;
+            count[level] = index;
+            if (index > 1) {
+              count = count.slice(0, level + 1);
+              symbols = symbols.slice(0, level + 1);
+            }
+          }
```
AndrewSouthpaw, line 370: The ordered numbering assumes contiguous heading levels. `count` is indexed by `level = rank - lowestRank`, so a document that skips a level (e.g. an `#` directly followed by `###`, which is common) leaves a hole in the array, and `count.join('.')` turns the hole into an empty segment.

```
# Top
### Deep
→ "1. [Top]", "    1..1. [Deep]"
```

lib/transform.js
```diff
@@ -5,6 +5,11 @@ var anchor = require("anchor-markdown-header"),
   contentGenerator = require("./content-generation"),
   md = require("@textlint/markdown-to-ast");
 
+const M = ["", "M", "MM", "MMM"];
+const C = ["", "C", "CC", "CCC", "CD", "D", "DC", "DCC", "DCCC", "CM"];
+const X = ["", "X", "XX", "XXX", "XL", "L", "LX", "LXX", "LXXX", "XC"];
+const I = ["", "I", "II", "III", "IV", "V", "VI", "VII", "VIII", "IX"];
```
AndrewSouthpaw: nit: `M`/`C`/`X`/`I` are single-letter names at module scope, not great, let's give them descriptive names (`ROMAN_THOUSANDS`, `ROMAN_HUNDREDS`, …).

doctoc.js
```diff
@@ -163,6 +163,13 @@ else if (indentWidth === undefined) { indentWidth = (mode === 'bitbucket.org' ||
 var minLines = argv['document-lines-min'] || 0;
 if (isNaN(minLines)) { log.error('Document min lines: ' + minLines + ' is not a number'), printUsageAndExit(true); }
 
+var format = argv['toc-list-format'] || 'unordered';
+if (format && format != 'ordered' && format != 'unordered') { log.error('TOC List format: ' + format + ' is not valid'), printUsageAndExit(true); }
+var ordered = format === 'ordered';
+var style = argv['toc-list-style'];
+var symbol = entryPrefix.split(',') || ['-'];
```
AndrewSouthpaw: Two things:
1. `entryPrefix.split(',') || ['-']` — `String.prototype.split` always returns a (truthy) array, so the `|| ['-']` fallback is dead. `entryPrefix` is also already defaulted to `'-'` at lines 135-137.
2. This value feeds `items.symbol` (singular), but `transform.js` only ever reads `items.symbols` (plural).

doctoc.js
```diff
@@ -163,6 +163,13 @@ else if (indentWidth === undefined) { indentWidth = (mode === 'bitbucket.org' ||
...
+if (ordered && style && style != 'number' && style != 'uppercase' && style != 'lowercase' && style != 'roman') { log.error('TOC List style: ' + style + ' is not valid'), printUsageAndExit(true); }
```
AndrewSouthpaw: The style is only validated/applied when `ordered` is true. Passing `--toc-list-style roman` without `--toc-list-format ordered` is silently ignored, let's include that here.

test/transform-list.js
```diff
@@ -0,0 +1,126 @@
...
+        '      1.1.2.1. [Main Usage](#main-usage)',
+        '  1.2. [Some More](#some-more)',
+        ''
+    ].join('\n'),
+    'TOC is correctly formatted'
+  )
+  t.end()
+});
+
+test('\nUppercase ToC ', function (t) {
```
AndrewSouthpaw: The names and styles are swapped between these two tests:
- "Uppercase ToC" (here) uses `style: 'lowercase'` and asserts lowercase output (`a.`, `a.a.` …).
- "Lowercase ToC" (line 56) uses `style: 'uppercase'` and asserts uppercase output (`A.`, `A.A.` …).

README.md
```diff
@@ -243,6 +243,26 @@ By default,
...
 - The min items is set to 1
 
+##### Format
+
+Use the `--toc-list-format` option to control the format of the toc list, i.e. `doctoc --toc-list-format unordered`. 
+
+The supported options are:
+
+ - `ordered`: the list items are numbered rather than usinh a symbol with the style controlled via the `--toc-list-style` option.
+ - `unordered`: the list items use a symnol. This is the default behaviour.
```
AndrewSouthpaw: Typos:
- 252: "rather than usinh a symbol" → "using a symbol"
- 253: "the list items use a symnol" → "use a symbol"

README.md
```diff
@@ -243,6 +243,26 @@ By default,
...
+##### Style
+
+Use the `--toc-list-style` option to control the style of the ordered toc list, i.e. `doctoc --toc-list-style roman`. 
+
+The supported options are:
+
+ - `lowercase`: the list items are identified using lowercase letters ie `a.c.a`.
+ - `number`: the list items are identified using numbers letters ie `1.3.1`. This is the default behaviour.
+ - `roman`: the list items are identified using roman numerals ie `I.III.I`.
+ - `uppercase`: the list items are identified using lowercase letters ie `A.C.A`.
```
AndrewSouthpaw: Typos:
- "identified using numbers letters" → "using numbers" (drop "letters"), or "using numerals"
- `uppercase` is described as "identified using lowercase letters" → should be "uppercase letters"

AndrewSouthpaw (changes requested):

lib/transform.js
thompson-tomo: ~~Actually it is used for tests & if someone was to use the transform module directly. If we don't have it nearly every single test will need to be changed or the defaulting is added at each usage of options.~~

Ahh, I see bad merge.

lib/transform.js
thompson-tomo: That is true but it would also report a markdown linting violation. I will see if there is a nice way to handle it.

doctoc.js
thompson-tomo: The readme makes it clear the default is unordered so I don't see an issue with requiring the user to consistently set ordered if they want an ordered list. They can then optionally set the style of the list.

Have added a check to log this scenario.

lib/transform.js
thompson-tomo: Done

lib/transform.js
thompson-tomo: done

thompson-tomo:
@AndrewSouthpaw all feedback has been addressed.

doctoc.js
```diff
@@ -169,6 +166,19 @@ if (isNaN(minLines)) { log.error('Document min lines: ' + minLines + ' is not a
 var location = argv['toc-location'] || 'top';
 if (location != 'top' && location != 'before') { log.error('Location specified is not valid: ' + location), printUsageAndExit(true); }
 
+var format = argv['toc-list-format'] || 'unordered';
+if (format && format != 'ordered' && format != 'unordered') { log.error('TOC List format: ' + format + ' is not valid'), printUsageAndExit(true); }
+var ordered = format === 'ordered';
+var style = argv['toc-list-style'];
+if (ordered && style && style != 'number' && style != 'uppercase' && style != 'lowercase' && style != 'roman') { log.error('TOC List style: ' + style + ' is not valid.'), printUsageAndExit(true); }
+else if (!ordered && style) { log.error('TOC List style: ' + style + ' should not be specified for unordered list.'), printUsageAndExit(true); }
+else if (ordered && style) { log.error('TOC List style: ' + style + ' should not be specified for unordered list.'), printUsageAndExit(true); }
```
AndrewSouthpaw: `--toc-list-format ordered --toc-list-style number` (and every other valid style) exit with an error. Seems like a copy-paste mistake?

lib/transform.js
```diff
@@ -5,6 +5,13 @@ var anchor = require("anchor-markdown-header"),
   contentGenerator = require("./content-generation"),
   md = require("@textlint/markdown-to-ast");
 
+const ROMMAN_NUMERALS = {
+  THOUSANDS: ["", "M", "MM", "MMM"],
+  HUNDREDS: ["", "C", "CC", "CCC", "CD", "D", "DC", "DCC", "DCCC", "CM"],
+  TENS: ["", "X", "XX", "XXX", "XL", "L", "LX", "LXX", "LXXX", "XC"],
+  ONES: ["", "I", "II", "III", "IV", "V", "VI", "VII", "VIII", "IX"]
+};
```
AndrewSouthpaw: nit: `ROMMAN_NUMERALS` → `ROMAN_NUMERALS` (double M).

README.md
```diff
@@ -259,6 +259,26 @@ By default,
...
+ - `ordered`: the list items are numbered rather than using a symbol with the style controlled via the `--toc-list-style` option.
+ - `unordered`: the list items use a symbol. This is the default behaviour.
+
+##### Style
+
+Use the `--toc-list-style` option to control the style of the ordered toc list, i.e. `doctoc --toc-list-style roman`. 
+
+The supported options are:
+
+ - `lowercase`: the list items are identified using lowercase letters ie `a.c.a`.
```
AndrewSouthpaw, line 277: Still has "numbers letters" — should be just "numbers" (drop "letters"), or "numerals".

AndrewSouthpaw (changes requested):

thompson-tomo:
All feedback has been implemented.

AndrewSouthpaw (approved):
Thanks for going through all the iterations here! Sorry it took a while to get through it.

thompson-tomo:
No worries at all. 🙂

PR #359, opened 2026-06-09
feat: Excude headings

### LINKED ISSUE #213: Feature request: add flag to look for TOC content before `<!-- START -->`
I've been converting some docs to MD that have TOC content before the TOC itself. e.g. `## Preface`. Currently,  only generates TOC content for elements found after the ```<!-- START -->```. It would be nice if the  was a _doctoc_ flag would override this behavior and parse the entire document. 

Naturally, this might require adding additional features to exclude referencing the TOC itself, like IGNORE flags.:

```<!-- TOC-IGNORE-START -->```
```<!-- TOC-IGNORE-END -->```

### PR DESCRIPTION
**Pull request by thompson-tomo**
Closes: #213 

This enables headings to be excluded from toc by adding the exclude tag directly before the heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,8 @@ In order to ignore a specific file when running `doctoc` on an entire directory,
 
 ### Exclude Heading
 
-In order to exclude a specific heading when running doctoc, just add `<!-- DOCTOC EXCLUDE -->` on a dedicated line directly before the heading which you wish to exclude. The exclude tag can not be nested within another html block and you should ensure that there is a blank line after the heading to avoid the parser considering the subsequent content as a part of the heading node.
+In order to exclude a specific heading when running doctoc, just add `<!-- DOCTOC EXCLUDE -->` on a dedicated line directly before the heading which you wish to exclude. 
+The exclude tag can not be nested within another html block and you should ensure that there is a blank line after your heading to ensure the heading will be excluded.
 
 ### Update existing doctoc TOCs effortlessly
 
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ by github or other sites via a command line flag.
   - [Configuring logging level](#configuring-logging-level)
   - [Adding toc to all files in a directory and sub directories](#adding-toc-to-all-files-in-a-directory-and-sub-directories)
   - [Ignoring individual files](#ignoring-individual-files)
+  - [Exclude Heading](#exclude-heading)
   - [Update existing doctoc TOCs effortlessly](#update-existing-doctoc-tocs-effortlessly)
   - [Adding toc to individual files](#adding-toc-to-individual-files)
     - [Examples](#examples)
@@ -241,6 +242,10 @@ specified](#using-doctoc-to-generate-links-compatible-with-other-sites).
 ### Ignoring individual files
 
 In order to ignore a specific file when running `doctoc` on an entire directory, just add `<!-- DOCTOC SKIP -->` to the top of the file you wish to ignore.
+
+### Exclude Heading
+
+In order to exclude a specific heading when running doctoc, just add <!-- DOCTOC EXCLUDE --> on a dedicated line directly before the heading you wish to ignore.
 
 ### Update existing doctoc TOCs effortlessly
 
--- a/README.md
+++ b/README.md
@@ -300,7 +300,19 @@ In order to ignore a specific file when running `doctoc` on an entire directory,
 ### Exclude Heading
 
 In order to exclude a specific heading when running doctoc, just add `<!-- DOCTOC EXCLUDE -->` on a dedicated line directly before the heading which you wish to exclude. 
-The exclude tag can not be nested within another html block and you should ensure that there is a blank line after your heading to ensure the heading will be excluded.
+
+An example of a compliant document with html headers is shown below with the toc only containg heading 2:
+
+```markdown
+<!-- DOCTOC EXCLUDE -->
+<h2>Heading 1</h2>
+
+<h2>Heading 2</h2>
+```
+
+The key requirements being:
+- exclude tag can not be nested within another html block and is on its own dedicated line
+- there is a blank line after your heading
 
 ### Update existing doctoc TOCs effortlessly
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/transform.js [hunk 1]
```diff
@@ -99,8 +99,11 @@ function getMarkdownHeaders (lines, start, maxHeaderLevel, minHeaderLevel) {
     return html.indexOf('<h', 3) === -1;
   }
 
+  var excludeText = contentGenerator.excludeTag(syntax);
+
   return lines
-    .children.filter(function (x) {
+    .children.filter(function (x, index, nodes) {
+      if (index > 0 && nodes[index - 1].raw === excludeText) return false;
```
AndrewSouthpaw: Exact string match on `.raw` is fragile here — a trailing space after the comment tag (e.g. `<!-- DOCTOC EXCLUDE --> `) causes the parser to produce `.raw = "<!-- DOCTOC EXCLUDE --> "` which won't match `excludeTag()`. Use `.trim`

lib/transform.js
```diff
@@ -266,7 +269,7 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
 
   var tocPosition = tocs.positions[0];
   var tocableStart = !processAll ? tocPosition.end.range[1] : 0;
-  var headers = getMarkdownHeaders(docNodes, tocableStart, maxHeaderLevel, minHeaderLevel)
+  var headers = getMarkdownHeaders(docNodes, tocableStart, maxHeaderLevel, minHeaderLevel, syntax)
     .concat(getHtmlHeaders(docNodes, tocableStart, maxHeaderLevelHtml, minHeaderLevel));
```
AndrewSouthpaw: The exclude logic only lives in `getMarkdownHeaders` — `getHtmlHeaders` has no exclude handling. So `<!-- DOCTOC EXCLUDE -->` before an `<h2>Method One</h2>` does nothing; the heading still appears in the TOC.

Either the exclude check needs to also apply to `getHtmlHeaders`, or this should be documented as markdown-only.

README.md
```diff
@@ -242,6 +243,10 @@ specified](#using-doctoc-to-generate-links-compatible-with-other-sites).
 
 In order to ignore a specific file when running `doctoc` on an entire directory, just add `<!-- DOCTOC SKIP -->` to the top of the file you wish to ignore.
 
+### Exclude Heading
+
+In order to exclude a specific heading when running doctoc, just add <!-- DOCTOC EXCLUDE --> on a dedicated line directly before the heading you wish to ignore.
```
AndrewSouthpaw: This renders the comment as invisible HTML — the user sees "just add on a dedicated line" with the tag completely hidden. Wrap it in backticks like the SKIP example above:

```suggestion
In order to exclude a specific heading when running doctoc, just add `<!-- DOCTOC EXCLUDE -->` on a dedicated line directly before the heading you wish to ignore.
```

AndrewSouthpaw (changes requested):

lib/transform.js
thompson-tomo: Actually the example of using <hX> tags should be supported provided they are not nested in html, will add a test and improve doc.

thompson-tomo: Test is now passing. Turns out main issue was the missing empty line hence have updated doc.

lib/get-html-headers.js
```diff
@@ -24,10 +24,21 @@ function addLinenos(lines, headers) {
...
+var go = (module.exports = function (nodes, start, maxHeaderLevel, minHeaderLevel, nested) {
+  function validMdHtmlHeader (html, minHeaderLevel, maxHeaderLevel, nested) {
+    if (nested === undefined) return true;
+    if (html[0] !== '<' || html[1] !== 'h') return nested;
+    var position = html.indexOf('<h', 3);
+    if (nested ? position === -1 : position !== -1) return false;
+    const level = html[2];
+    if (!html.endsWith('</h' + level + '>')) return nested;
+    if (nested) return true;
+    return Number(level) >= minHeaderLevel && Number(level) <= maxHeaderLevel || 4;
```
AndrewSouthpaw: Operator precedence bug: `Number(level) <= maxHeaderLevel || 4` parses as `(... <= maxHeaderLevel) || 4`. 

```suggestion
    return Number(level) >= minHeaderLevel && Number(level) <= (maxHeaderLevel || 4);
```

README.md
```diff
@@ -296,6 +297,11 @@ specified](#renderers).
 
 In order to ignore a specific file when running `doctoc` on an entire directory, just add `<!-- DOCTOC SKIP -->` to the top of the file you wish to ignore.
 
+### Exclude Heading
+
+In order to exclude a specific heading when running doctoc, just add `<!-- DOCTOC EXCLUDE -->` on a dedicated line directly before the heading which you wish to exclude. 
+The exclude tag can not be nested within another html block and you should ensure that there is a blank line after your heading to ensure the heading will be excluded.
```
AndrewSouthpaw: IMO this is a bit hard to parse, and an example would go a long way here.

AndrewSouthpaw (changes requested):

README.md
thompson-tomo: Changes have been made. Let me know your thoughts.

README.md
```diff
@@ -296,6 +297,23 @@ specified](#renderers).
...
+```markdown
+<!-- DOCTOC EXCLUDE -->
+<h2>Heading 1</h2>
+
+<h2>Heading 2</h2>
+```
+
+The key requirements being:
+- exclude tag can not be nested within another html block and is on its own dedicated line
+- there is a blank line after your heading
```
AndrewSouthpaw:
```suggestion
An example of a compliant document with html headers is shown below with the toc only containing heading 2:

```markdown
<!-- DOCTOC EXCLUDE -->
<h2>This heading will not show up in the TOC</h2>

<h2>Heading 2</h2>
```

To exclude a heading:
- the exclude tag cannot be nested within another html block
- the exclude tag must be on its own dedicated line
- there is a blank line after your heading
```

AndrewSouthpaw (approved):

README.md
AndrewSouthpaw: LGTM, I made some tweaks to clarify language further in https://github.com/thlorenz/doctoc/pull/359/commits/5dd1941a9aea92404690eeb1758b52e8229ee9f0.

PR #360, opened 2026-06-13
Add support for inserting anchors

thompson-tomo:
Progresses: #174 
Closes: #218

This adds in the ability to insert anchors into the doc which is necessary for bit bucket server. However it has been in a generic manner enabling reuse.


thompson-tomo:
Blocked on https://github.com/thlorenz/anchor-markdown-header/pull/73

PR #363, opened 2026-06-19
feat: Detect comment syntax

thompson-tomo:
I noticed that the syntax option is not documented. While writing the doc and checking the implementation the opportunity to streamline the implementation was identified. This streamlining helps with the documentation.

The changes are:
- syntax defaults to being detected based on file extension
- Syntax options now match syntax type ie html, jsx with determination occurring during filesystem scanning.

the benefit is we can support mixed environments ie mdx & md within the same workspace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,8 @@ Supported options are:
 
 > [!NOTE]
 >
-> By supplying a syntax, the files to be processed will be filtered ie html syntax is used in md & markdown files.
+> By specifying a syntax, the file scanner will filter the files to be processed based on the syntax & extension.
+> For instance, html syntax would result in only md & markdown files being processed.
 
 #### Min. Document Lines
 
@@ -130,6 +131,7 @@ By default,
 > If your document contains images, those images will not be counted any different to how plain text is,
 > as doctoc is processing your document as a text processor.
 > This also means repeated newlines will also be considered should your document contain them.
+
 #### Renderers
 
 In order to add a table of contents whose links are compatible with other sites add the appropriate mode flag:
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ Supported options are:
 - **md**: an alias for html. Will be removed in v3.
 - **mdx**: an alias for jsx. Will be removed in v3.
 
+> [!NOTE]
+>
+> By supplying a syntax, the files to be processed will be filtered ie html syntax is used in md & markdown files.
+
 #### Min. Document Lines
 
 Use the `--document-lines-min` option to specify the minimum lines required to be in a document for the document to have a table of contents; e.g., `doctoc --document-lines-min 200 .`.
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ This is useful in CI environments where you want to check if your docs are up to
 
 You can print to stdout by using the `-s` or `--stdout` option.
 
-This option is only applicable when specifying a single filename which doctoc is to run on. If you are specifying a folder or multiple files, the dry run option should be used instead.
+This option is only applicable when specifying a single filename or a directory containing only 1 file which doctoc is to run on. If you are specifying a folder with multiple files or providing multiple inputs, the dry run option should be used instead.
 
 #### Update only
 
@@ -88,7 +88,7 @@ Use the `--syntax` option to specify the syntax used for comments in the documen
 
 Supported options are:
 
-- **jsx**: comments start with `{\*` and end with `*\}`.
+- **jsx**: comments start with `{/*` and end with `*/}`.
 - **html**: comments start with `<!--` and end with `-->`.
 - **md**: an alias for html. Will be removed in v3.
 - **mdx**: an alias for jsx. Will be removed in v3.
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ by github or other sites via a command line flag.
     - [Stdout](#stdout)
     - [Update only](#update-only)
   - [Document Options](#document-options)
+    - [Comments](#comments)
     - [Min. Document Lines](#min-document-lines)
     - [Renderers](#renderers)
   - [Source Options](#source-options)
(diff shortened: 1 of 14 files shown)


lib/file.js
```diff
@@ -47,5 +55,17 @@ function findRec(paths, syntax) {
 // Finds all markdown files in given directory and its sub-directories
 // @param {String  } dir - the absolute directory to search in
 exports.findMarkdownFiles = function (dir, syntax) {
-  return findRec([dir], syntax);
+  var target = cleanPath(dir);
+  var stat = fs.statSync(target);
+  if (stat.isDirectory()) {
+    log.debug('\nDocToccing "%s" and its sub directories', dir);
+    return findRec([target], syntax);
+  } else {
```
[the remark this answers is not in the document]
thompson-tomo: No the values are normalised before usage.

thompson-tomo:
@AndrewSouthpaw here is another cli option which wasn't documented and by documenting it i discovered it was confusing. 😂

lib/file.js
```diff
@@ -25,10 +31,12 @@ function findRec(paths, syntax) {
         stack.push(fullPath);
       }
 
-      if (fileInfo.isFile() && allowedExts.includes(path.extname(fileInfo.name))) {
+      if (fileInfo.isFile() && allowedExts.includes(path.extname(fileInfo.name)) &&
```
AndrewSouthpaw, line 36: When the user passes `--syntax md` (or `mdx`), `findRec` compares it against `syntaxMapping` values which are `'html'` / `'jsx'`. The check `syntax === syntaxMapping[ext]` becomes `"md" === "html"` → `false`, so every file is silently filtered out and nothing gets processed.

Let's add test coverage here too.

lib/content-generation.js
```diff
@@ -38,6 +38,9 @@ function compactPragma(syntax) {
 }
 
 function commentedBlock(syntax, content) {
+  if (syntax == "md") syntax = "html";
+  if (syntax == "mdx") syntax = "jsx";
```
AndrewSouthpaw: The md→html / mdx→jsx aliasing is now scattered across three files (here in `commentedBlock`, `escapedStartTag`, plus `transform.js`). Each call-site independently maps the aliases, and one was already missed (`file.js` filtering — see other comment).

Consider normalizing once at the entry point (either in `doctoc.js` before passing to `findMarkdownFiles`, or in a shared helper) so downstream code only ever sees `'html'` or `'jsx'`. That would let you delete these guards and the ones in `transform.js`.

lib/file.js
```diff
@@ -47,5 +55,17 @@ function findRec(paths, syntax) {
...
-  return findRec([dir], syntax);
+  var target = cleanPath(dir);
+  var stat = fs.statSync(target);
+  if (stat.isDirectory()) {
+    log.debug('\nDocToccing "%s" and its sub directories', dir);
+    return findRec([target], syntax);
+  } else {
+    log.debug('\nDocToccing single file "%s".', dir);
+    return [{
+      path: target,
```
AndrewSouthpaw, line 70: When the target is a single file, no syntax filtering is applied — the file is always returned regardless of `--syntax`. In directory mode, files are filtered by syntax. This means `doctoc file.md --syntax jsx` processes the file with JSX comments (wrong comment style for .md), but `doctoc dir/ --syntax jsx` correctly skips .md files. That seems unexpected.

doctoc.js
```diff
@@ -209,24 +199,15 @@ if (argv._.length > 1 && stdOut) {
...
 
 for (var i = 0; i < argv._.length; i++) {
-  var target = cleanPath(argv._[i]),
-    stat = fs.statSync(target);
+  var files = file.findMarkdownFiles(argv._[i], argv['syntax']);
 
-  if (stat.isDirectory() && stdOut) {
-    console.error('--stdout cannot be used to process a directory. Use --dryrun instead.');
+  if (files.length > 1 && stdOut) {
+    console.error('--stdout cannot be used to process a directory containing more than 1 file. Use --dryrun instead.');
```
AndrewSouthpaw: Minor wording fix:

```
'--stdout cannot be used when more than 1 markdown file is found. Use --dryrun instead.'
```

AndrewSouthpaw (commented):
Nice refactor overall — auto-detecting syntax from extensions and supporting mixed .md/.mdx workspaces is a solid improvement. A few issues below, one of which is a correctness bug.

AndrewSouthpaw:
One last thing: `test/file.js` tests pass despite the filtering bug

The tests in `test/file.js` hit the broken code path but pass anyway because of `[].every()`:

```javascript
findMarkdownFiles(fixturesDir, 'md').every((file) => file.path.endsWith(".md"))
```

`[].every(pred)` returns `true`, so it always passes.

Also the third test says "undefined syntax" but passes `'md'` — copy-paste error.

Can you clean these up while you're here?

lib/content-generation.js
thompson-tomo: Won't it be a breaking change if it is only normalized at entrypoint? Hence have normalized it using a helper.

lib/file.js
thompson-tomo: Hmm, this was the previous behaviour before refactor. I have updated the readme to improve clarity.

thompson-tomo:
@AndrewSouthpaw can you take another look as i think i have implemented all your feedback.

doctoc.js
```diff
@@ -2,22 +2,15 @@
 
 "use strict";
 
-var path = require("path"),
-  fs = require("fs"),
-  os = require("os"),
+const { normalizeSyntax } = require("./lib/utils/helpers");
```
AndrewSouthpaw: Unused import — `normalizeSyntax` is never called in this file. The normalization happens inside `findMarkdownFiles` and `transform` now, so this `require` can be removed.

README.md
```diff
@@ -87,14 +88,32 @@ This is useful in CI environments where you want to check if your docs are up to
...
+Supported options are:
+
+- **jsx**: comments start with `{/*` and end with `*/}`.
+- **html**: comments start with `<!--` and end with `-->`.
+- **md**: an alias for html. Will be removed in v3.
+- **mdx**: an alias for jsx. Will be removed in v3.
+
+> [!NOTE]
+>
+> By specifying a syntax, the file scanner will filter the files to be processed based on the syntax & extension.
```
AndrewSouthpaw, line 114: Missing `> ` prefix — this line renders outside the NOTE block. Should be:

```md
> For instance, html syntax would result in only md & markdown files being processed.
```

AndrewSouthpaw (commented):
Good progress — the `normalizeSyntax` helper fixes the filtering bug and centralization is cleaner. A few remaining items.

AndrewSouthpaw:
**test/file.js** — the `syntax=md` and `syntax=mdx` tests still use `.every()` without a length check. Same pattern that masked the original bug. Now that filtering works they won't be empty, but a future regression would silently pass again with zero results. The `undefined` syntax test was properly fixed — worth applying the same `length > 0` assertion to the other two.

PR #369, opened 2026-06-26
feat: modularise toc part generation with variable padding length

thompson-tomo:
This extracts the generation of the toc parts into the content generator module.

This extraction improves testability, functionality scoping & lays the foundation for enabling full control of padding before & after the content.

This approach enables control of the qty of padding before title etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,8 +241,6 @@ It is recommended to also include this argument when performing an update on an 
 
 Use the `--toc-title-padding-before` option to add padding line/s above the TOC which ensures formatters such as prettier will pass; e.g., `doctoc --toc-title-padding-before 1 .`
 
-NOTE: Currently it is only supported to add one line before the title.
-
 By default,
 
 - no padding is added above the table of contents title
--- a/lib/content-generation.js
+++ b/lib/content-generation.js
@@ -56,34 +56,33 @@ function escapedStartTag(syntax) {
 }
 
 function tocFooter(footerOptions) {
-  if (footerOptions?.content === undefined) return [];
-  var lines = Array(footerOptions?.padding?.before ?? 0).fill('');
-  lines.push(footerOptions.content);
-  lines.push(...Array(footerOptions?.padding?.after ?? 0).fill(''));
-  return lines;
+  return tocPart(footerOptions?.padding, footerOptions?.content);
 }
 
 function tocHeader(headerOptions) {
-  if (headerOptions?.content === undefined) return [];
-  var lines = Array(headerOptions?.padding?.before ?? 0).fill('');
-  lines.push(headerOptions.content);
-  lines.push(...Array(headerOptions?.padding?.after ?? 0).fill(''));
-  return lines;
+  return tocPart(headerOptions?.padding, headerOptions?.content);
 }
 
 function tocTitle(titleOptions, existingTitle) {
   if (titleOptions?.remove || (existingTitle === "" && !titleOptions?.content)) return [];
-  var lines = Array(titleOptions?.padding?.before ?? 0).fill('');
+  var content;
   if (titleOptions?.content){
-    lines.push(titleOptions.content);
+    content = titleOptions.content;
   }
   else if(existingTitle){
-    lines.push(existingTitle);
+    content = existingTitle;
   }
   else{
-    lines.push('**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*');
+    content = '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*';
   }
-  lines.push(...Array(titleOptions?.padding?.after ?? 0).fill(''));
+  return tocPart(titleOptions?.padding, content);
+}
+
+function tocPart(padding, content){
+  if (content === undefined) return [];
+  var lines = Array(padding?.before ?? 0).fill('');
+  lines.push(content);
+  lines.push(...Array(padding?.after ?? 0).fill(''));
   return lines;
 }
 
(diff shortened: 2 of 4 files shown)


AndrewSouthpaw:
Just a heads up, I’ve been dealing with a serious family medical situation
for the past six weeks, I’ll try to catch up on reviews in the next week or
two. Thanks for your patience.


--

thompson-tomo:
No worries at all @AndrewSouthpaw hope that everyone is on the mend. 🤞

README.md
```diff
@@ -241,8 +241,6 @@ It is recommended to also include this argument when performing an update on an
 
 Use the `--toc-title-padding-before` option to add padding line/s above the TOC which ensures formatters such as prettier will pass; e.g., `doctoc --toc-title-padding-before 1 .`
 
-NOTE: Currently it is only supported to add one line before the title.
```
AndrewSouthpaw, line 244: The CLI guard at `doctoc.js:145` is still in place:

```
$ doctoc --toc-title-padding-before 2 README.md
Padding before title: 2 is not currently supported as greater than 1
```

So for anyone using the CLI nothing has changed. The guard needs to come out in this PR too if you no longer want this caveat.

lib/content-generation.js
```diff
@@ -55,10 +55,44 @@ function escapedStartTag(syntax) {
...
+  }
+  else{
+    content = '**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*';
+  }
+  return tocPart(titleOptions?.padding, content);
+}
+
+function tocPart(padding, content){
+  if (content === undefined) return [];
+  var lines = Array(Number(padding?.before ?? 0)).fill('');
```
AndrewSouthpaw, line 83: The CLI needs validation on non-negative integers, e.g.:

```
$ doctoc --toc-title-padding-before 0.5 README.md
RangeError: Invalid array length
    at tocPart (lib/content-generation.js:83:15)
```

Similar for `1.5`.

Let's add some CLI arg validation?

lib/transform.js
```diff
@@ -338,16 +330,15 @@ exports = module.exports = function transform(content, mode, maxHeaderLevel, min
...
     var lowestRank = allHeaders.reduce((min, h) => Math.min(min, h.rank), Infinity);
     
     var indentation = options.toc.items.indentation.style === 'tab' ? '\t' : ' ';
 
-    if (options?.toc?.header?.content) { tocLines.push(options.toc.header.content); }
-    if (padTitle && inferredTitle) { tocLines.push(''); }
-    if (inferredTitle) { tocLines.push(inferredTitle); }
+    tocLines.push(...contentGenerator.tocHeader(options.toc.header));
+    tocLines.push(...contentGenerator.tocTitle(options.toc.title, inferredTitle));
     tocLines.push('');
```
AndrewSouthpaw, line 342: `padding.after` stacks on top of this hardcoded blank rather than replacing it, so the new knob is effectively off by one — `{ toc: { title: { padding: { after: 1 } } } }` produces two blank lines:

```
**Table of Contents**  *generated with [DocToc](...)*
                              ← padding.after
                              ← this line
- [A](#a)
```

Similar to footer.

Can we add tests to cover this?

lib/content-generation.js
```diff
@@ -55,10 +55,44 @@ function escapedStartTag(syntax) {
   return commentTypes[syntax].escapedStart;
 }
 
+function tocFooter(footerOptions) {
+  return tocPart(footerOptions?.padding, footerOptions?.content);
+}
+
+function tocHeader(headerOptions) {
+  return tocPart(headerOptions?.padding, headerOptions?.content);
```
AndrewSouthpaw, line 63: `header.padding` and `footer.padding` are read here but nothing ever sets them — no flag in `doctoc.js`, no mention in the README, no test. Same for `title.padding.after`. Only `title.padding.before` has a way in.

Either wire up the flags and docs in this PR (which is what the title suggests), or leave `padding` off header/footer until there's a way to reach it. As-is it reads like a feature but is dead for every CLI user.

lib/content-generation.js
```diff
@@ -55,10 +55,44 @@ function escapedStartTag(syntax) {
...
+}
+
+function tocTitle(titleOptions, existingTitle) {
+  if (titleOptions?.remove || (existingTitle === "" && !titleOptions?.content)) return [];
```
AndrewSouthpaw, line 67: This creates a with `determineTitle` that nothing spells out: `undefined` means "nothing known, use the default", `""` means "the existing TOC deliberately had no title, keep it that way", and a non-empty string means "preserve this one". Maybe worth a simple object, like:

```
const titleSetting: {
  default: undefined,
  existingTocNoTitle: "",
  preserveTocTitle: "...",
}
```

Or at least some comments.

lib/content-generation.js
```diff
@@ -55,10 +55,44 @@ function escapedStartTag(syntax) {
...
 module.exports = {
   commentedBlock,
   escapedStartTag,
   pragmaMarkers,
   skipTag,
-  excludeTag
+  excludeTag,
+  tocFooter,
+  tocHeader,
+  tocTitle
```
AndrewSouthpaw, line 97: General thought: add direct unit tests for `tocTitle`'s fallback chain (explicit content → existing title → default → removed).

AndrewSouthpaw (changes requested):

README.md
thompson-tomo: Done.

lib/transform.js
thompson-tomo: That is expected. The hard coded padding is for the items. It is no different to enabling padding after header & before title.

